PR #11, opened 2015-04-10
added check for NA in get_cid()

Dahaniel:


EDiLD:
Thanks Daniel! I suspect, this will be also the case for `get_csid()` and other functions? Need to check... #12

Dahaniel:
Probably, but as I only used get_cid() and cir_query() I only recognized it there.

EDiLD:
Fixed also in other functions. Now there is also documentation for bulk processing.

Dahaniel:
great!

Issue #2, opened 2015-04-02
FatePointer


EDiLD:
http://esc.syrres.com/fatepointer/search.asp

EDiLD:
#16 implements this partly (PHYSPROP Database).

PR #16, opened 2015-08-19
Dev physprop

EDiLD:


EDiLD:
Implement physprop (see also issue https://github.com/ropensci/webchem/issues/2).

PR #24, opened 2015-08-26
Correct Alan Woods first name everywhere

jranke:
This PR renames the function allanwood() to alanwood() and updates the package to reflect this.

Should an alias allanwood() be added for backward compatibility? Personally, I think it would be better not to.


EDiLD:
Oops :flushed:, 

Thanks Johannes for fixing this typo.

Issue #118, opened 2016-09-20
Add ping_*() functions to check if APIs are running


EDiLD:
From a reviewer:

> Some of the man pages I looked at indicated that the query might fail if the API is unavailable. It might be useful to have a helper method to test the availability of the remote resource before the query is run.

- [ ] allanwood
- [ ] chemid
- [ ] chemspider API
- [x] chemspider web page
- [ ] cir
- [ ] cts
- [ ] etox
- [ ] opsin
- [x] pan
- [ ] physprop
- [x] pubchem
- [ ] wikidata

Krisselack:
Using cir_query on a char-vector sometimes resulted for me in the following error: 
"Error in curl::curl_fetch_memory(url, handle = handle) : 
  Timeout was reached"
The problem was, that no NA was returned in that case and the sequence got messed up.

EDiLD:
Thx for reporting this. I couldn't reproduce this and it *should* return NA in case of a timeout.

It basically implemented like this:

```r
library(httr)
h <- try(GET("http://httpbin.org/delay/3", timeout(1)))
if(inherits(h, 'try-error')) NA
```

Can you provide a reproducible example?

Krisselack:
I don't know how reproducible it is, but after running this script it occured a few times. 
To save time, now I query for unique entries... strangely, without that error. 

[chems.txt](https://github.com/ropensci/webchem/files/541872/chems.txt)
chems <- read.csv("chems.txt",stringsAsFactors = FALSE)
chems[which(chems=="PAK 16")] <- "PAK_16"
chem_namen <- cir_query(chems,'names',first=TRUE)

stitam:
I see that we currently export `ping_cs()`, `ping_pan()`, `ping_pubchem()`, `ping_pubchem_pw()`. I was wondering how this will look like when we implement `ping_*()` functions for all web services and also add 10 more web services;). Also @Aariq introduced the `is.down()` functionality for tests in #248 which is very similar to `ping()`. Maybe we could export only one `ping()` function which calls various non-exported ping functions internally and use this function in the tests as well? Would this make things more simple/consistent?

Aariq:
A simple change that I can imagine being helpful is changing it to just `ping_service(service)` where service is the prefix for the functions for that service. Then, when a new service is added, we just need to add a URL to a `switch()` statement or something like that.

stitam:
[...]

PR #119, opened 2016-09-26
Make get_cid work again

jranke:
mhm, now the original function from the master branch and the pubchem tests works again, so it seems this pull request can be deleted - maybe the API was temporarily unavailable?

In this case it would be great to be able to give a corresponding error message...

### DIFF
--- a/R/pubchem.R
+++ b/R/pubchem.R
@@ -43,18 +43,15 @@ get_cid <- function(query, from = 'name', first = FALSE, verbose = TRUE, arg = N
 
   foo <- function(query, from, first, verbose, ...){
     prolog <- 'http://pubchem.ncbi.nlm.nih.gov/rest/pug'
-    input <- paste0('/compound/', from)
+    input <- paste0('/compound/', from, "/")
     output <- '/cids/JSON'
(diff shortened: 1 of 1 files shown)


EDiLD:
> In this case it would be great to be able to give a corresponding error message...

Yes, this is on my TODO-List, see issue #118.
I suspect it was just a temporary downtime...

EDiLD:
@jranke Just saw your [chents project](https://github.com/jranke/chents), where you use this.
Looks like a very interesting project and I wondered want the future plans for this are...

Also a nice idea to call RDKit through PyhtonInR! I currently use rrdkit, but it is not easy to get working and probably not as flexible. I might also adopt this approach...

jranke:
@EDiLD yes, I just revived the chents package to some extent. I work on it when I think I can use it in my projects. In general, the idea is to use it for comparative risk assessment of chemicals, but that's a long way to go. BTW, I did not know about rrdkit. Wow, work on it started already 2014... Using the python bindings works fine, and setting up PythonInR should be a minor thing. Somehow it is handled different on Windows and Linux, but everything is documented well.

Issue #127, opened 2017-04-05
cs_prop throws error with some CSIDs


Aariq:
cs_prop() seems to not work with certain CSIDs

For example `cs_prop(16105)` yields:

> NAs introduced by coercionError in data.frame(prop, value_pred, unit_pred, source_pred, value_exp,  :   arguments imply differing number of rows: 7, 5, 6

Other examples of CSIDs that give a similar error:
- 9912
- 13876539
- 7889
- 11885
- 73261
- 28985

Aariq:
Some of the problems are caused by missing values in the EPI properties and dealt with those, but there are also situations where there are multiple values for some properties.  E.g. csid 7812 has two water solubility estimates.

EDiLD:
Thank you for your valuable contribution and extensive checking!
I will have a look at it and merge the pull request by the end of the next week.

Aariq:
FYI it doesn't totally fix the problem.  Occasionally there is more than
one entry for a particular property and that still causes it to fail.  A
possible temporary fix is adding an option to only get ACD values (which
work fine), but I'm not familiar enough with writing functions to know how
to do that off the top of my head.

-Eric

EDiLD:
Commit c6d8089 fixes this bug and passes all tests with the above mentioned CSID.

`cs_prop` is currently a bit messy...
- [ ] Currently experimental BP and MP are ignored
- [ ] Other fields returned by EPISuite are also ignored and should be incorporated.

EDiLD:
In master branch now.

PR #128, opened 2017-04-05
fix cs_prop() issue #127

Aariq:
The problem was with the EPIsuite data.  Some compounds don’t have
values for all types of data so occasionally results from gsub() were
vectors of zero length.  I added an if statement after every gsub to
check that something actually got added to the vector.  This is maybe
not the most elegant solution, and is probably overkill because not all
properties are subject to the possibility of missing values (I think
only experimental values are missing, not predicted values).


EDiLD:
Made a different fix, see issue #127.

PR #136, opened 2017-11-23
Fixed stripping of . from extracted numbers

stanstrup:
Right now numbers are not extracted corrected.
For examples in pp_query.

`extr_num("Melting Pt : -44.6 deg C")` strips the "." and gives -446.

My fix keeps "."'s. It will fail if there is a dot in another place though.
`extr_num("Melting Pt : -44.6 deg C.")` gives NA.

Alternatively you could use `readr::parse_number` that also handles that.


stanstrup:
Hmm. This strips the - instead.

stanstrup:
OK fixed that.

EDiLD:
Cherry-picked https://github.com/ropensci/webchem/pull/136/commits/d2201f82c40ce98270336ae30d65022b0433bcb5 , added tests & merged into master.  Closes #136 .

Can you separate PRs regarding issues (easier for me to check & merge). Will now look at #138 and #139.

Thank you for your welcome contributions!

EDiLD:
Cherry-picked all suggestions.

PR #140, opened 2018-01-11
bugfix for is.smiles

allaway:
Hello - I recently ran into an issue with this function using the current CRAN version of webchem, but I believe this issue still exists in the master branch here. Apologies if I am supposed to file a pull request against a different branch than master. 

is.smiles previously returned TRUE for any string regardless of validity. I suspect this was due to changes in rcdk::parse.smiles. 

This change should correctly return TRUE for valid SMILES, and FALSE for un-parseable SMILES.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -261,12 +261,14 @@ is.smiles <- function(x, verbose = TRUE) {
     stop('Cannot handle multiple input strings.')
   }
   out <- try(rcdk::parse.smiles(x), silent = TRUE)
-  if (inherits(out, 'try-error') | is.na(out)) {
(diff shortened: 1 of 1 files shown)


EDiLD:
Thanks for your contribution! I'll check and test ASAP (might take few days).
In fact, i had to turn off the tests for is. smiles() , as java is a bit b****y 😞 

Could also use rdkit the test if it parses (could be implemented via rrdkit, or (better) via reticulate.

EDiLD:
Thanks for your contribution!
 Incorporated you changes in `master` and fixed tests and continuous integration (rcdk installation failed).

PR #144, opened 2018-02-18
pc_synonyms multiple compounds per request

damondpham:
In this new version, `pc_synonyms` will submit `CHUNKSIZE` compounds per request, instead of one per request. This should be faster.

This is my first pull request so I apologize for any mistakes; hopefully this is helpful!

### DIFF
--- a/R/pubchem.R
+++ b/R/pubchem.R
@@ -242,7 +242,7 @@ pc_synonyms <- function(query, from = 'name', interactive = 0, verbose = TRUE, C
     )
     
     if (inherits(cont, "try-error")) {
-      if(length(query) > 1){
+      if (length(query) > 1) {
         warning('Problem with web service encountered... Trying chemicals individually.')
         cont <- lapply(query, foo, from=from, verbose=verbose)
       } else {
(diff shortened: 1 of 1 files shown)


Aariq:
I get an error using this version of pc_synonyms "Error in names(object) <- nm : 
  'names' attribute [11] must be the same length as the vector [2]"

EDiLD:
@damondpham Thanks for your contribution! I am currently busy with work and can't handle this at the moment, earliest in two weeks (volunteers, welcome - I'll grant full access to responsible contributors).

Can you adress the points raised by @Aariq (and @Aariq can you provide an reproducible example?).

Note that I have unit tests defined for most functions. You can find them in the `tests` branch. Simple merge your changes into this branch and run the tests and link results here  (I separated tests from master, as I don't want the tests be run on each push - being nice to the servers).

damondpham:
I fixed the error, and within two weeks I can run those tests.

I realized one downside with submitting multiple chemicals per request is that, if the request fails because of one chemical, all chemicals will get a NA result. I think this can be fixed by re-trying in the case of failure, but submitting one chemical per request instead of all as a group.

EDiLD:
>  and within two weeks I can run those tests.

Great! 

You can also run only the pc_* tests locally...

damondpham:
I'm actually not sure how to run the tests and link my results here nicely. What kind of files would I submit?

I did run the four test cases for pc_synonyms locally, and they all passed. This is the only function I changed.

stitam:
Thank you @damondpham for your contribution!

Given that:
* The checks are still failing
* The source repo has probably been deleted
* There has been no activity on this PR in the past 1.5 years

I think it is safe to assume that this PR has been abandoned. I am sorry we couldn't merge it, I really hope we will see more contributions from you in the future! I am closing this pull request.

PR #148, opened 2018-04-10
Fixed errors when data is duplicated on CS

stanstrup:
More protection against value duplication. See CSID 7688 where everything is duplicated. I just inserted the save_val function everywhere.


EDiLD:
Thanks for you contribution! I'll add tests and documentation.

stanstrup:
No problem. Happy for the package. chemspider parsing is like a game of whack a mole.

PR #150, opened 2018-05-07
Working around for physprop when compound is present but there are no properties

stanstrup:
Example: http://esc.syrres.com/fatepointer/webprop.asp?CAS=75706126


EDiLD:
Thanks for your contribution! I created also an issue #151 and will do some residual fixes (NEWS, tests) after merging.

Issue #159, opened 2018-11-26
as.cas() doesn't handle NAs


Aariq:
as.cas() should always return `NA` when run on `NA`. Currently it gives an error:
```r
as.cas(NA)
##Error in if (nsep != 2) { : missing value where TRUE/FALSE needed
##In addition: Warning message:
##In as.cas(NA) : Some elements of x cannot be converted to CAS numbers

x = c(58082, 123456, "hexenol", NA)
as.cas(x)
##Checksum is not correct! 5 vs. 6
##Less than 2 hyphens in string.
##Error in if (nsep != 2) { : missing value where TRUE/FALSE needed
##In addition: Warning message:
##In as.cas(x) : Some elements of x cannot be converted to CAS numbers
```

Aariq:
Fixed in PR #161

Issue #158, opened 2018-11-17
`cir_query()` returns sodium ion for `NA` in character vectors


Aariq:
When `cir_query()` is passed a character vector including `NA`s, it thinks you are asking for the chemical element sodium.

```r
cir_query(NA) #gives an error, as it should
## Error in strsplit(URL, "") : non-character argument

test <- c(NA, "64-17-5")
cir_query(test, "names", first=TRUE)

##                <NA>             64-17-5 
## "sodium(+1) cation"           "ethanol" 
```

Aariq:
Fixed in PR #161

PR #161, opened 2018-12-17
addresses issue #158 and #159

Aariq:
Fixes two issues, #158 where `cir_query()` was mistaking NAs for sodium (Na) and #159 where as.cas(NA) was throwing an error.  Both functions now return NAs when given NAs, this makes them easier to use on large vectors that might contain NAs.

```r
# insert reprex here
```

PR task list:
- [ x] Update NEWS
- [x ] Add tests (if appropriate)
- [x ] Check package passed


EDiLD:
Thank you.

PR #169, opened 2019-04-13
Fixes #168

### LINKED ISSUE #168: PAN query fails
Error in ``pan_query`` results. 

```r
> pan_query('2,4-D')
Error in strsplit(out[["Chemical Name and matching synonym"]], "\\n"):
non-character argument
```

gjgetzinger:
String substitution for pan_query output replaced with ``\n`` with ``""`` (e.g., ``Chemical Nameand matching synonym``) Later code indexed output with name containing a space (eg., ``Chemical Name and matching synonym``). 

Replaced empty character with space in 
```r
names(out) <- gsub('\\n', ' ', names(out))
```

```r
webchem::pan_query("2,4-D")
```

### DIFF
--- a/R/pan.R
+++ b/R/pan.R
@@ -113,7 +113,7 @@ pan_query <- function(query, match = c('best', 'all', 'first'), verbose = TRUE, 
     out <- as.list(ttt)
     # clean
     out$`Detailed Info` <- NULL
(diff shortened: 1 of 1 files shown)


EDiLD:
Thank you for the fix!

Issue #157, opened 2018-11-17
add interactive mode to all resolvers that return multiple names


Aariq:
Having an interactive mode for choosing metabolite names based on another identifier is key for my workflow.  It is implemented in `pc_synonyms`, but it would be great to have it implemented in other functions that return names (or even multiple identifiers), such as `cir_query()`

This is a relatively easy addition.  Just add `interactive = 0` to the function arguments and add this code to the function:

```r
    if (interactive > 0 && length(out) > 1) {
      pick <- menu(out[seq_len(interactive)], graphics = FALSE, 
        "Select one:")
      out <- out[pick]
    }
```

I know current maintainers don't have time, I'm putting this out there in case anyone wants to implement this before I get a chance.

Aariq:
I'm now thinking that this would be best implemented as a standalone helper function that any function which returns a vector of names can call. What do you all think?  I've personally found this pretty helpful, but maybe it won't get that much use?

PR #175, opened 2019-07-12
new "choices" argument to replace "first" and "interactive"

Aariq:
Begins to address issue #157 by adding a utility `chooser(x, choices)` that returns only the first result if `choices = 1`, and provides the user with an interactive menu to choose a result if `choices` > 1 or "all".  If choices = NULL, then the vector is returned unchanged.  

I've updated three functions to use this (pc_synonym, cir_query, and cts_convert), although there might be others where it makes sense.  Basically I think it is useful whenever a function returns many possible matches and the user might only want one.  "first" never made sense for pc_synonym because the first result is often (always?) a CAS number, not a chemical name.

I also made some minor updates to documentation to encourage contributors to use this utility in new functions when appropriate.

I think there are other options for the best way to do this, so if you don't like it I'm happy to have a discussion about the best way to implement this (or not).

BTW, the examples for `cts_convert()` don't work currently because there is something wrong with cts when converting from InChIKey.  Try converting from chemical name if you want to test this out.

PR task list:
- [x] Update NEWS
- [x] Check package passed


R/utils.R
```diff
@@ -366,3 +366,42 @@ as.cas <- function(x){
...
+    }
+    if(choices == "all") { #then give all of x as possible choices
+      pick <- menu(x, graphics = FALSE, 'Select one:')
+      out <- x[pick]
+    }
+    if(choices == 1) {
+      out <- x[1]
+    }
+    if(is.numeric(choices) & choices > 1){
+      pick <- menu(x[seq_len(choices)], graphics = FALSE, 'Select one:')
```
stitam:
```suggestion
      pick <- menu(head(x, choices)], graphics = FALSE, 'Select one:')
```

Aariq: Thanks for the feedback.  This makes sense.

stitam:
Hi Aariq,

Thanks for the update! I like the new chooser() function, and given that you said it is important for your workflow, I think it should be included in the package. What I am uncertain about is including calls to chooser() in other functions. I feel it would be more efficient if we just called chooser() on the output of other functions whenever we needed the interactivity. That way we wouldn't have to incorporate new lines of code to our other functions, and we could also keep our functions as simple as possible. What do you think?

Aariq:
I mostly just like having the option to return a vector instead of a list and `first = TRUE` never made much sense for a synonym function since you're probably not just looking for any old synonym that's arbitrarily at the top of the list. So when I wrote `pc_synonyms`, it originally had an `interactive` argument instead of `first=TRUE`. In this pull request, I just thought I'd move the code to do the interactive choosing out of `pc_synonyms()`and into the utils.R file so others could re-use it (there are some other minor improvements over the old code in `pc_synonyms` as well).  Any function that has a `first = ` argument can use `choices` instead, and it _reduces_ the lines of code in that function since it just runs the results through `chooser()` (`first = TRUE` has the same effect as `choices = 1`).  But I understand that it might be seen as overkill for functions where you probably do want either the entire vector of results or an arbitrary first result.

Here's an example of how I like to use the `choices` argument in `pc_synonyms()` (previously `interactive=` ):

```r
library(tidyverse)
library(webchem)
df <- tibble(cas = c("123-35-3", "108-88-3", "122-78-1"))

df %>% 
  mutate(syn = pc_synonyms(cas, choices = 10))
```
Without the `choices` argument, `pc_synonyms` returns vectors of different lengths for each row and it makes a list column, which is a pain to deal with.

Of course there are other ways I could do this by mapping the `chooser()` function to the list that `pc_synonyms(df$cas)` returns, like you suggested.

Aariq:
I guess what I'm saying is I'm happy to revert the changes to `cir_query` and `cts_convert` and just keep this about an improvement to `pc_synonym` and creation of `chooser()` if you'd prefer.

stitam:
Let's keep the changes as you suggested. I'm merging the pull request.

PR #187, opened 2019-10-07
Fix class definition orders, closes #185

### LINKED ISSUE #185: Class assignments
etox_basic() returns an object of classes:
```
attr(,"class")
[1] "list"       "etox_basic"
```
whereas the list should start with the more specific class.

### LINKED ISSUE #186: Add print methods for the classes returned in the package

stitam:


stitam:
I found that functions in extractor.R are indeed using these class definitions.

Issue #180, opened 2019-08-15
as.cas() fails on correctly formatted CAS numbers!


Aariq:
Whoops!  `as.cas()` should probably first check if the cas number is already formatted correctly with `is.cas()`.  It should also maybe then try removing all non-numeric characters before moving on.

```r
library(webchem)
is.cas("24903-95-5")
#> [1] TRUE
as.cas("24903-95-5")
#> Warning in as.cas("24903-95-5"): Some elements of x cannot be converted to
#> CAS numbers
#> Less than 2 hyphens in string.
#> [1] NA
```

PR #188, opened 2019-11-01
Fix for #180

Aariq:
This fixes a problem where `as.cas()` would fail on correctly formatted CAS numbers.  Also slightly changes the behavior to remove a blanket warning ("some elements couldn't be converted to CAS numbers") and instead just show the warnings passed by `is.cas()`.

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -354,6 +354,8 @@ as.cas <- function(x){
   format.cas <- function(x){
     if(is.na(x)) {
       return(NA)
+    } else if (suppressMessages(is.cas(x))) {
+      return(x)
     } else {
       parsed <- gsub("([0-9]+)([0-9]{2})([0-9]{1})", '\\1-\\2-\\3', x)
       pass <- is.cas(parsed)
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -354,16 +354,16 @@ as.cas <- function(x){
...
       pass <- is.cas(parsed)
       out <- ifelse(pass, parsed, NA)
       return(out)
     }
   }
-  if(any(!sapply(x, grepl, pattern= "^\\d+$"))){
-    warning("Some elements of x cannot be converted to CAS numbers")
-  }
+
   sapply(x, format.cas, USE.NAMES = FALSE)
```
stitam: I know this is not new code, but I don't see a scenario where `sapply()` is being executed, all of the `if()` routes lead to a `return()` statement. Do you think this line could be removed from the function?

Aariq: `sapply()` is always being executed.  The if statements are part of the `format.cas()` function.  Using `sapply()` makes the function vectorized. I'm open to suggestions if there is a better way to do this though.

stitam: I just overlooked something, this is perfect the way it is, thanks!

PR #192, opened 2020-01-08
Minor updates before next release

stitam:
I found the list of contributors in README.Rmd redundant with the list within GitHub Insights so I removed it. Added myself as maintainer as CRAN will need a single contact that can be reached during the submission process. Let me know if you disagree. Anything else you would recommend before I submit the new release?


DESCRIPTION
```diff
@@ -15,8 +15,8 @@ Authors@R: c(person("Eduard", "Szöcs", role = c("aut", "cre"),
     person("Eric R", "Scott", role = "ctb"),
     person("Jan", "Stanstrup", role = "ctb"),
     person("Robert", "Allaway", role = "ctb"),
```
andschar: can you add my name as a contributor?
person("Andreas", "Scharmüller", role = "ctb"),

Aariq:
I did just realize that some of the help files are slightly messed up because I thought using markdown in roxygen comments was turned on, but it's not.  It's mostly just instances of backticks instead of `\code{}`, which is not a big deal.

stitam:
I have changed bacticks to `\code{}` and rerun `devtools::document()`. The latter updated other parts of the documentation as well.

Issue #190, opened 2020-01-04
cs_check_key() function to access API token stored in .Rprofile or .Renviron


Aariq:
I've been looking at how other packages deal with API keys, and realized that I have several stored in my own .Renviron file.  Some, like the `rorcid` package or `usethis`, just provide instructions in the documentation on how to edit the .Rprofile or .Renviron file.  Others, like `rtweet` provide functions to write to the .Renviron file.  All three of these packages have an (sometimes not exported) function to access the API token stored either in .Renviron or .Rprofile that is then called when a token is not supplied to a function that needs it.  This seems like a good practice for us to implement for any databases that require API keys (currently only chemspider?). 

For example, here's the `check_key()` function from `rorcid`:

```
function () 
{
  x <- Sys.getenv("ORCID_TOKEN", "")
  if (x == "") {
    x <- getOption("orcid_token", "")
  }
  if (x == "") 
    NULL
  else x
}
```

Aariq:
- [ ] write `check_key` function(s)
- [ ] change all `cs_*` functions to use the `check_key` function by default
- [ ] remove `apikey` from examples
- [ ] remove API keys from tests and update code in tests
- [ ] remove API keys from README.Rmd
- [ ] get README to build on Travis

PR #195, opened 2020-01-09
Access API tokens from .Renviron or .Rprofile

Aariq:
## Pull Request

Related to #190 

- [x] Added functions `cs_check_key()` and `cs_check_token()` to retrieve ChemSpider API key and secure token from .Renviron or .Rprofile
- [x] Updated all `cs_*` functions to use  these functions by default to get the API key.
- [x] Removed private API keys and secure tokens from all documentation and tests
- [x] Updated NEWS
- [x] Check package passed (although I'm unsure how it will work with CI

I also turned on using markdown with `roxygen2` comments and double-checked that this didn't cause any problems in the documentation.


stitam:
Thanks @Aariq this is really good stuff. Do you want to merge this before v0.5.0?

Aariq:
No, go ahead and deploy v0.5.0.  This can be part of the next version.  I
think there might still be some kinks to work out.

Aariq:
Looks like maybe this is failing on Travis because it's using the development version of R and the installation of `stringi` fails for that version.  I'm going to try editing travis.yml to test it on the release version to see if it passes.

stitam:
Thanks @Aariq, I think this PR is almost ready to be merged, let's merge it before v0.5.0! `cs_extcompinfo()` uses the old API, and I haven't found an alternative in the new, so currently this functionality cannot be fixed. Therefore, I deprecated the function, but maybe it would be better to put it to defunct? Regardless, I don't think it would need `cs_check_token()`. What do you think? Have you seen any other ChemSpider functions that use tokens?

Aariq:
I think I found all the instances of `apikey` and changed them to use the new `cs_check_token`.  I think just keep `cs_extcompinfo()` as deprecated for now.

R/chemspider.R
```diff
@@ -1,3 +1,61 @@
...
+#' }
+cs_check_token <- function() {
+  x <- Sys.getenv("CHEMSPIDER_TOKEN", "")
+  if (x == "") {
+    x <- getOption("chemspider_token", "")
+  }
+  if (x == "")
+    stop("no security token stored for ChemSpider.  See ?cs_check_token() for details")
+  else x
+}
```
stitam: After the API change at ChemSpider, they are not using tokens anymore, but they use API keys for all their APIs. `cs_extcompinfo()` is currently not working and cannot be fixed, but even if they introduce another API in the future that allows us to fix it, we would use the same `cs_check_key()` function you propose for all other ChemSpider functions. Because of this, I think it would be better not to include `cs_check_token()` at all, only `cs_check_key()`, and also erase all reference to "tokens" from anywhere but the already deprecated `cs_extcompinfo()` function. What do you think?

Aariq: Ah, thanks.  I wasn't sure what the difference between a key and a token was.  I'll make your suggested change.

Aariq:
Not quite sure why so many .Rd files unrelated to my changes got reformatted, but it doesn't change the rendered output as far as I can tell.  We might have different versions of `roxygen` or something?

stitam:
It is possible. I have `7.0.2`, the repo is currently using `6.1.1`. Since there has been a major release, I think it would be better to upgrade to the latest version. Can you please make this last commit before I merge your PR?

Issue #197, opened 2020-01-13
Physprop is not reachable anymore


andschar:
The PHYSPROP database is not reachable anymore for some time now. Hence `pp_query()` doesn't work. Unfortunately, there is not much information about the Physprop database in general. Maybe they put it in the [EPI Suite](https://www.epa.gov/tsca-screening-tools/epi-suitetm-estimation-program-interface)? Does anybody know anything about it?

This URL would normally lead to a result:
```r
http://esc.syrres.com/fatepointer/webprop.asp?CAS=50000
```

stitam:
Thank you for opening this issue @andschar!

I think you are right, PHYSPROP is in EPA's EPI Suite (another source [saying the same thing](https://www.srcinc.com/what-we-do/environmental/scientific-databases.html)). The good news is that ChemSpider already uses the EPI Suite, when you search [their website](https://www.chemspider.com/Default.aspx) for a compound, you can navigate to "Properties" and then "Predicted - EPISuite". This is good news because part of that information is accessed by `cs_prop()`. The bad news is that not everything is captured by the function. I will open another issue to fix `cs_prop()`.

If you find more information about PHYSPROP, please post it here and let's discuss whether the function can be fixed or it should be removed.

andschar:
There was never much information to find about Physprop. There isn't even a project website resp. an address to contact somebody. Or am I missing something? Should we just remove it and re-install it in case it will again be reachable? Leaving it in webchem rather confuses user imho.

stitam:
I think SRS owns PHYSPROP, I contacted them on Monday but I haven't received a reply yet. I agree it is confusing, we shouldn't release v0.5.0 without solving this. The question is whether to fix, deprecate or defunct. I'd say let's wait till the weekend and if we don't get an answer by Saturday, defunct. @andschar would you like to do the commit?

andschar:
Ok, sounds good to me. Can do that.

stitam:
I haven't received a reply from SRS, so feel free to defunct `pp_query()` and remove PHYSPROP from the documentation. Thanks!

stitam:
I have received a reply from SRS saying that PHYSPROP is no longer available outside of EPA’s EPISUITE program.

PR #205, opened 2020-01-26
Defunct pp_query()

andschar:
I have defunct pp_query() since the API is not active anymore. I hope I did everything correctly.
I realized that ci_query() also provides Physprop data and I have changed the note accordingly.
I have also updated the man pages and fixed a typo in CONTRIBUTING.
This resolves issue #197


R/chemid.R
```diff
@@ -18,8 +18,8 @@
 #' inchi (vector), inchikey (vector), smiles(vector), toxicity (data.frame),
 #' physprop (data.frame) and source_url.
 #'
-#' @note The data of the entry \code{pp_query} is identical to the result returned
-#' by \code{\link{pp_query}}.
+#' @note The data of the entry physprop was identical to the result returned
+#' by \code{\link{pp_query} in webchem versions < 0.5}.
```
stitam: If we remove `pp_query()` then I think this note is not informative anymore. Maybe we could remove the whole note?

andschar: I was undecided whether to remove it or not. Ok, let's better remove the whole note. Will do.

andschar:
Ok, sorry. I wasn't aware of the roxygen2 update. It seems it changed the formatting.

Issue #207, opened 2020-01-31
Upgrade get_csid() to accept more input types


stitam:
At the moment `get_csid()` only accepts compound names.

stitam:
I think this issue can be closed.

Issue #206, opened 2020-01-28
Convert chemcal formula to CID or compound name


MjelleLab:
Is it possible to convert chemical formula such as "C47H93N2O6P" to CID or compound name?

Aariq:
Because there are many chemicals with that formula, I'm not sure that we have a function that does this.

Aariq:
Actually, I take that back.  I think you *should* be able to use `get_csid("C47H93N2O6P")` to do this, but right now it seems to only return an empty list.  `get_csid("H2O")` works, for example, but returns IDs for compounds other than water.  Any ideas @stitam ?

andschar:
I had a deeper look at the function and I realized that we are querying the name-endpoint of the API in `get_csid()`:
https://developer.rsc.org/compounds-v1/apis/post/filter/name
H2O seems to be a name too for Chemspider that's why a result is returned. However there's a specific formula-endpoint for that, which also returns a result for the OP's formula: C47H93N2O6P
https://developer.rsc.org/compounds-v1/apis/post/filter/formula
The best would probably be, to use an argument to specify which API endpoint to query, meaning `type = 'formula'` or similar.

andschar:
I could work that in for `get_csid()` if you want. Also for the v0.5.0 release, since I can spare some time this week.

biocoreexpert:
Hi, that would be great!

stitam:
Thanks @biocoreexpert for raising this issue also @andschar for the detailed answer about `get_csid()`. True, the function currently implements the name endpoint, and implementing the formula endpoint would allow us to create csid from formula. Would this help? However, with the ChemSpider APIs we can only convert csid to inchi, inchikey, smiles or mol. I tried the csid -> name conversion with CTS but it doesn't seem to work (not even on the website).

In general, `get_cid()` (note: cid, not csid) accepts formulae. I tried it with H2O and C3H8NO5P (glyphosate) and it works, although for some reason it returns only one result whereas the website returns more, we'll look into this. However, for C47H93N2O6P it doesn't return anything, whereas the website returns 60 results. On the other hand, if I take the CID from the website, `pc_prop()` will return an object that contains the IUPAC name of the compound. Hope this helps @biocoreexpert, if you have more compounds, try `get_cid()`, it might work for some. Thanks again for bringing this up!

stitam:
[...]

Issue #199, opened 2020-01-22
remove documentation for non-exported functions


Aariq:
In previous contributions, I didn't realize that even if a function isn't exported, if it has roxygen comments it gets documented.  I think it would be better if this wasn't the case.  There is an @noRd tag that I can add to any internal functions that removes the documentation visible to the user while keeping the roxygen comments for developers.

Related to #194

Issue #193, opened 2020-01-08
get_*() consistency discussion


Aariq:
Currently `webchem` has `get_` functions for 4 databases to retrieve database-specific IDs that can then be used as input to other functions.  These all have different interfaces.
```
get_cid(query, from = "name", first = FALSE, verbose = TRUE,
  arg = NULL, ...)

get_csid(query, apikey, control = cs_control())

get_etoxid(query, match = c("best", "all", "first", "ask", "na"),
  verbose = TRUE)

get_wdid(query, language = "en", match = c("best", "first", "all",
  "ask", "na"), verbose = TRUE)
```
Let's use this as a starting point to discuss consistency of the API across databases. Personally, my favorite arguments that I think should be in all of the `get_` functions are:

- `query` for obvious reasons should always be the first argument.
- the `from` argument let's you specify whether the input is a name, inchikey, etc. and the documentation specifies what kinds of queries are valid for that particular database.
- the `first` argument lets you get a vector out, even if multiple IDs are returned.  Vectors of IDs are integral for using in other `webchem` functions that return dataframes. I like the `match` argument in `get_wdid` too, but how often are you really going to use "ask" for ID numbers?  How are you supposed to know which is the right one beforehand?
- `...`, although I think they should pass arguments to an underlying control function like `cs_control()` (that is, I don't think `get_cid` should have both `arg` and `...`)
- database-specific arguments like `language` for `get_wdid` or `apikey` for `get_csid` should go at the end (but before `...`)

Anyway, happy to hear what y'all think.  Hopefully this will be an easy-ish jumping off point for making the API more consistent across databases.

Aariq:
related: #191, #157

Aariq:
Also related: #206

andschar:
We could also turn `chebi_lite_entity()` into a `get_` function, since it is basically one, returning possible matches as e.g. `get_cid()`.

andschar:
I also think that `from = `  and `first = ` are important arguments.

Aariq:
Another (maybe bad) idea is to get rid of `get_` functions entirely, or just use them internally.  Unless we think users will want to get database specific IDs without any other data, why use them?  Instead, functions like `cs_compinfo()` could accept multiple types of queries rather than just CSIDs, and return a dataframe with the query and the CSID as columns.

andschar:
I think the `get_` functions are not bad at all. For the last big project, I was first collecting different identifiers with these functions and then used them to query additional data.

[...]

PR #208, opened 2020-02-02
Upgrade get_csid() to accept more input types, etc.

stitam:
* `get_csid()` now accepts `name`, `formula`, `inchi`, `inchikey`, `smiles`. Default is `name`.
* `get_csid()` now returns a data frame, and not a named list. `csid` is the first column.
* `cs_convert()` now returns vectors.
* Style was updated as recommended by `lintr::lint()`
* There was a typo in `cs_check_key()`, `option->options`, also added some more detail.

PR task list:
- [x] Update NEWS
- [x] Check package passed


stitam:
This PR is related to Issue #191, Issue #193, Issue #206, Issue #207.

stitam:
I think it would be better not to export the `cs_*_csid()` functions, only the top level functions that use them, `get_csid()` and `cs_convert()`. I think it helps with ease of use, as it provides a single function to get a csid from all sorts of query types and another to convert between all sorts of id types. Are you suggesting that we shouldn't create `roxygen2` documentation for non exported functions? I don't have a strong opinion on this. The documentation was meant for future developers. Do we expose a documentation if we don't export the function itself?

Aariq:
If the documentation is for developers only, then I wouldn't create the .Rd file.  You can add a @noRd to keep the roxygen2 comments in the .R file, but not generate a help page.  I've been meaning to do this for a couple of other functions as well. #199

stitam:
You are right, we have already discussed this, thanks! I will update this PR.

PR #210, opened 2020-02-10
copyediting documentation

Aariq:
This is mostly minor copyediting of documentation.  It also removes .Rd files for non-exported functions (#199), and makes one code change to `is.cas()`.  I noticed that `is.cas()` would just ignore its second argument if it was something other than `FALSE` so I changed `if(verbose)` to `if(isTRUE(verbose))`

PR task list:
- [ x] Update NEWS
- [x] Check package passed


R/utils.R
```diff
@@ -382,7 +381,7 @@ as.cas <- function(x){
 #' @importFrom utils menu
 #' @importFrom utils head
 #' @return a character vector of length 1
-#' @export
```
stitam: I am OK with not exporting `chooser()`, but I added a comment related to this a few days ago in #193, have you seen it?

Aariq: I did see that.  I'm still thinking about the response.  I figured this version of the function should not be exported—if we do end up exporting it, I assume the functionality and/or documentation will have to change a bit anyway.

stitam: That's true. You're right, let's not export this version of `chooser()`.

Issue #183, opened 2019-09-20
Function to get citations for databases


Aariq:
I imagine in addition to citing webchem, people using this package may also want to (or be technically required to) cite the databases that webchem accesses.  It might be a good idea to have a function that returns citations for each database, similar to `citation()`.  Maybe you could use the function prefix to get the citation.

```
webchem_cite("cs")
#>To cite ChemSpider in publications use:
#>    Ayers, M. (2012), "ChemSpider: The Free Chemical Database", Reference Reviews, Vol. 26 No. 7, pp. 45-46. https://doi.org/10.1108/09504121211271059
#> A BibTex entry for LaTeX users is:
#> <bibtex citation>
```

Aariq:
then `citation("webchem")` could reference this function saying "in addition to citing `webchem`, cite any relevant databases you've accessed. Get citations using the `webchem_cite()` function.  See `?webchem_cite` for more details."

And now that I think of it, it might be more consistent for functions to be like `cs_cite`, `fn_cite`, `nist_cite` etc.  These could be wrappers to `webchem_cite` and all have the same help page.

PR #211, opened 2020-02-12
Update function documentation with Terms and Conditions

stitam:
I added links to third party information that regulate how data can be accessed and used.

PR task list:
- [ ] Update NEWS - no need I think
- [x] Check package passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .Ruserdata
 .vscode*
 .Renviron
+.Rprofile
 
 README_files*
 README.html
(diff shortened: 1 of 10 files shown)


Aariq:
Since this is related, do you think you could also add something to the citation for webchem to remind users to check the help files for any databases they used for information on attribution?

to add to CITATION:

```
citFooter("In addition to citing webchem in publications, please check attribution guidelines for any data sources you accessed.  Links to documentation can be found in help files for webchem functions")
```

Then I think I'd feel comfortable closing #183 with this PR

PR #221, opened 2020-03-03
Minor changes before v0.5.0

stitam:
These are the last issues I had to fix before v.0.5.0 was approved for CRAN.

PR task list:
- [x] Update NEWS
- [x] Check package passed


andschar:
Sorry, forgot the `webchem:::` ;)

stitam:
Actually you didn't:) This is what the automatic test returned during the submission process:

"There are ::: calls to the package's namespace in its code. A package almost never needs to use ::: for its own objects:"

And so I had to remove the `:::` call to pass the test. I didn't know if there was a policy on this or not now it's clear:)

PR #223, opened 2020-03-17
Fix get_cid to return NA when query is NA and style update

stitam:
Most of the PR is just style update according to `lintr::lint()`. The real difference is in line 59 of pubchem.R. `get_cid()` returned the Pubchem ID of sodium when the query was `NA`.

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,7 @@ MINOR IMPROVEMENTS
 
 BUG FIXES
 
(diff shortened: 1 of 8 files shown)


Aariq:
Nice!  We should probably be careful about proper handling of NA for *all* get_ functions, not just because of the sodium issue (which has shown up before), but because we want them to be properly vectorized.

Issue #224, opened 2020-03-17
Functions handle NAs in input incorrectly


stitam:
* `aw_query(NA)` returns http error 404.
* `get_chebiid(NA)` returns sodium containing compounds.
* `ci_query(NA)` returns `NA`, but the output structure is different from a valid query.
* `get_csid(NA)` and all other ChemSpider functions return client error: (400) bad request.

These are just the first few, I believe most functions in the package are affected. Since a lot of webchem functions are vectorised, I think it is important that these functions always return the same output structure even if the input is invalid. I think we have to find a robust solution and implement it systematically in all of our affected functions.

Aariq:
I had a feeling this one was more pervasive.  Something we should keep in mind for future contributions as well.

stitam:
Thanks for bringing this up in PR #223! The issue came up in an example for the paper where `get_cid()` and `get_csid()` were affected. I cannot think of a better alternative than to edit each function one by one. What do you think? I will edit the contributing guide to remind the contributors of this issue.

Aariq:
Just need to edit each function to have something like:

```r
if(is.na(query){
  return(NA)
}
```
And add tests to make sure it still returns the expected data type (i.e. a tibble).

stitam:
Yes, something like this. However, if a function returns e.g. a list of 8 elements, then it has to return the same structure when the input is `NA`.

andschar:
I agree that the same __structure__ should be returned, but not necessarily with the same number of elements. So if a function returns a `data.frame` with 4 columns, it's enough to return `data.frame(NA)` because `data.table::rbindlist()` has the `fill = TRUE` argument and `dplyr::bind_rows()` fills values automatically. Except for the case of only NA results.

```r
l = list(iris[1:10, ], data.frame(Species = NA))
dplyr::bind_rows(l)
data.table::rbindlist(l, fill = TRUE)
```
So, a structure like you suggest in PR #225 in the chebi.R file is in my opinion not necessary. What do you think?

andschar:
Yes, it's a rather complex data structure that is returned from ChEBI (i.e. a ontology - a data graph) which is too difficult, or better an overkill to squash into __one__ data.frame in my opinion. So I think there's no reason to bind it in general (be it for the user or us webchem devs). `chebi_comp_entity()` mostly returns a list of lists, but yes some list elements are again lists.  
[...]

PR #225, opened 2020-03-19
Alternatives for handling NA inputs

stitam:
This PR is meant to contribute to the discussion on issue #224.

The interesing part is in `chebi.R`. The issue is that if the input is invalid, e.g. `"balloon"` (unknown chemical) or `NA`, the structure of the output will be different, which makes vectorisation difficult. As a reprex, `chebi_comp_entity(c("CHEBI:89301","CHEBI:89302","balloon",NA))` will return a nicely formatted list for the first two entries and something else for the last two. This makes `sapply(test, function(x) x[["properties"]]$inchikey)` return NULL-s, and unable to coerse the output into a character vector.

My approach for handling invalid inputs is to return an empty object with the same structure but all `NA`-s. I tried two alternatives: in `get_chebiid()` I defined the structure twice within the function. In `chebi_comp_entity()` this would have been distrubingly complex, so I created a separate, non exported function, which generates an object with the correct structure, but all `NA`-s inside, and called this function in lines 226 and 251. Forcing to return an empty object for invalid inputs allowed `sapply()` to extract inchikeys and coerce the output to a character vector.

I think the second approach is more elegant and less error prone. What do you think about the these empty objects? Can you suggest other solutions?

PR task list:
- [x] Update NEWS
- [x] Check package passed


R/alanwood.R
```diff
@@ -183,6 +181,8 @@ build_aw_idx <- function() {
   aw_idx$linknames <- ln
   attr(aw_idx, "date") <- Sys.Date()
 
+  # do not delete this line
+  # occasionally the data file should be updated in the package
```
Aariq: I wonder if there is some way to do this automatically whenever the package is built?

R/chebi.R
```diff
@@ -116,23 +120,30 @@ get_chebiid <- function(query,
...
-        return(out[ which.max(out$searchscore), ])
+        out <- out[with(out, order(searchscore, decreasing = TRUE)), ]
+        return(out[which.max(out$searchscore), ])
       }
       if (match == "ask") {
         matched <- chooser(out$chebiid, 'all')
-        return(out[ out$chebiid == matched, ])
+        return(out[out$chebiid == matched, ])
       }
       if (match == 'na') {
```
Aariq: I don't understand why match would be "na". I see that it's an option in line 86, but I don't get why the user would ever choose that as an option.  Is there some condition that causes it to be set to "na" that I'm not seeing?

R/chebi.R
```diff
@@ -72,8 +72,12 @@ get_chebiid <- function(query,
                         ...) {
```
Aariq: Since you're working on this function, you might change `match = "all"` in line 68 to `match = c("all", "best", "ask")` and change line 87 to `match <- match.arg(match)`.

Aariq: BTW, I wouldn't do this for `from` since there are a lot of possible choices.  Better to keep the options for that argument in the documentation only I think.

R/alanwood.R
stitam: We could [run a pre-built script when building the package](https://stackoverflow.com/questions/23723870/running-a-pre-build-script-when-building-an-r-package), but that wouldn't be automatic either because we would have to call the script manually. Also I don't think we can automate a function that writes on someone's hard drive, even if that someone is a developer. Maybe a better practice would be to remove the data file from the package completely, and build the index whenever `aw_query()` is first used in a session. We could use `tempdir()` to store the index for the session I believe this is accepted practice. What do you think?

R/chebi.R
stitam: Thanks for the suggestion, I updated the `match` arguments and `match.arg()`.

R/chebi.R
stitam: I can only guess. If I set `verbose = FALSE`, I may not know if there were multiple hits for a query. In that case an `NA` line in the output could be helpful for identifying queries that require further attention. The `match` argument in most functions has this option, so for CI and consistency reasons I'd rather not remove this option for now.

R/alanwood.R
Aariq: I like this suggestion, but it's probably something to handle in a separate pull request.  Might as well keep it in data/aw_idx.rda for now

R/chebi.R
Aariq: So according to the `get_wdid()` documentation, `match = "na"` is supposed to return `NA` if there is more than one result.  I.e., if your query results in more than one wikidata ID, then return NA.  Is that the behavior here?

R/chebi.R
```diff
@@ -340,4 +350,103 @@ rbind.named.fill <- function(x) {
...
+}
+
+#' Create empty list for chebi_comp_entity
+#'
+#' The function creates a list with identical internal structure as a
+#' chebi_comp_entity() output element. This is used in chebi_comp_entity() to
+#' provide a valid output for invalid inputs and allow vectorisation.
+#' @noRd
+#' @autor Tamas Stirling, \email{stirling.tamas@@gmail.com}
+empty_chebi_comp_entity <- function() {
```
andschar: Do we need this since it adds considerably more code? See #224 .

R/alanwood.R
stitam: Thanks for bringing this up @Aariq, I'll open a separate issue for this topic.

R/chebi.R
stitam: Yes, that is the intended behavior. `get_chebiid("triclosan")` returns three results, while `get_chebiid("triclosan", match = "na")` returns NA, so I think the functionality works as intended. Does this answer your question?:)

andschar: Yes, @Aariq  that's the behavior. When I updated the funciton, I basically looked at `get_wdid()` and implemented the same options here.

stitam: Also `get_etoxid()` has the same behavior.

Aariq: But if there is only one ID it returns the ID?  I just didn't see in the code where a length 1 result would skip this if statement.  It just looked to me like it would always return `NA` as long as `match = "na"`.  If it has the correct behavior, then I'm fine with this.

stitam: Actually that was the case earlier so thanks for the hint:) I added line 123 a few hours ago, now it should be fine.

R/chebi.R
stitam: Maybe not, suggested an alternative in #224.

Issue #227, opened 2020-03-24
Empty Etox tables


andschar:
The following call returns an error because the table doesn't contain any data: 
```r
etox_basic(99015)
```
https://webetox.uba.de/webETOX/public/basics/stoff.do?language=en&id=99015
The code breaks [here](https://github.com/ropensci/webchem/blob/dad1b7468afbc86a3ee830265c8fea92ed60893f/R/etox.R#L193).
Maybe we could add something like 
```r
if (inherits(tabs, 'try-error')) {
  message('ID not found! Returning NA.\n')
  return(data.frame(NA))
}
```
Not sure about it though.

stitam:
I think this could work. However, this topic is related to issue #224, and PR #225. An invalid input to the etox functions should return a structure that allows subsequent vectorised extraction of elements, e.g. cas number in this case, so `return(data.frame(NA))` might not be sufficient.

andschar:
Do we already know how many extractor functions we want to have? Currently there are cas, inchikey and smiles.

stitam:
I think the more the better:) I am currently working on some extractors for alanwood and chebi (only cas, inchikey, smiles)

Issue #218, opened 2020-02-26
Standardizing what is returned by functions


Aariq:
There is a huge amount of variation in what object type and structure is returned by `get_*` and database querying functions in webchem.  Some of this variation might be necessary (e.g. because the data contained in a database is varied), but I think it would be ideal for functions to return data.frames whenever possible.

Here's a quick and dirty overview of what formats are currently returned by functions in webchem:

# Get ID functions

- `get_chebiid`: list of dataframes (one per query)
- `get_cid`: list of vectors
- `get_csid`: dataframe (one row per query?)
- `get_etoxid`: dataframe with column for query.
- `get_wdid`: with `match = "all"`,  list of a single vector with attribute.  For `match = "first"`, a data frame

# query database functions

- `aw_query`: nested list (one list per query)
- `chebi_comp_entity`: complex nested list (each compound has a nested list with some - elements being dataframes and others being lists ofvectors)
- `chebi_lite_entity`: list of dataframes (even with a single query)
- `ci_query`: complex nested list
- `cs_compinfo`: a dataframe with one row per query
- `cts_compinfo`: complex list
- `etox_basic`: complex list
- `etox_targets`, `etox_tests`: list with a dataframe and a source URL(character vector) - nested under each query
- `fn_percept`: named vector (one element per query)
- `nist_ri`: dataframe with column for query
- `pan_query`: nested list.  For each query, the list seems like all length 1 vectors. - this should be a data frame.
- `pc_prop`: data.frame
- `pc_synonyms`: list of character vectors (one vector per query) unless choices != NULL, - then a dataframe with queries as rownames
- `srs_query`: list of dataframes (one data frame per query) which include list columns

# Ideas 

Something that might help us think through this is if someone came up with a "homework" problem that needs to be solved with multiple data sources and ID translations.  Then we can each share our solutions (e.g. as vignettes in our own forks of webchem) and do some thinking on what would result in the least friction for the most users. The results of this could be a vignette for our package and/or supplemental for the webchem publication.

Whenever possible, I think we should:

1. keep output consistent regardless of arguments or number of queries
2. prioritize dataframes whenever possible (except maybe functions that **always** return one element per query, which might be more usable as named vectors).

stitam:
[...]

PR #229, opened 2020-03-25
fixed etox_basic

andschar:
fixed `etox_basic()` issue addressed in #227 
changed output of etox_basic
added `cas.etox_basic()` method to R/extractors.R

PR task list:
- [ ] Update NEWS
- [x] Check package passed


R/etox.R
```diff
@@ -190,24 +189,39 @@ etox_basic <- function(id, verbose = TRUE) {
...
+                          stringsAsFactors = FALSE)
+    chem_group <- as.data.frame(t(group), stringsAsFactors = FALSE)
+    names(chem_group) <- chem_group[1, ]
+    chem_group[1, ] <- TRUE
+    out <- list(data = data,
+                chemical_group = chem_group,
+                synonyms = syn)
 
-    out <- list(cas = cas, ec = ec, gsbl = gsbl, synonyms = syns,
-                source_url = qurl)
```
stitam: I'm OK with your changes in the output structure, my only arguement against them is CI:) Can you please explain why the new structure is better? Also `source_url` is missing from the returned elements, I presume that was not intentional?

R/etox.R
```diff
@@ -190,24 +189,39 @@ etox_basic <- function(id, verbose = TRUE) {
       message('ID not found! Returning NA.\n')
       return(NA)
     }
-    tabs <- html_table(tt, fill = TRUE)
+    tabs <- try(html_table(tt, fill = TRUE), silent = TRUE)
```
stitam: When I run `etox_basic(99015)` it return `NA` which is great, but it also returns two warning messages. As it turns out `silent = TRUE` turns off the error messages, but it does not turn off the warning messages. Maybe you could turn those off as well with `suppressWarnings()`?

R/etox.R
stitam: Also if the structure will be changed, then the example in line 170 will also have to be changed.

R/etox.R
andschar: I also realized that and thought of `supressWarnings()`. Will implement.

R/etox.R
andschar: (1) will implement `source_url`, forgot it
(2) will change the example
(3) the old structure was a list of `data.frame`s and `vector`s, meaning various types. I changed them in light of harmonizing the data output to `list` of `data.frame`s.

andschar:
Updated the PR. Why does CI concern you? Its checks passed. I'm not very familiar with CI I have to admit.

R/etox.R
stitam: As I understand this is a different from those discussed in #218, #224, #225 so far, but I like the idea of grouping together similar items like IDs, and returning a list of data.frames/tibbles for each query when possible. I would rather leave the structure unchanged for now because of CI, but open a discussion on the topic. What do you think?

Aariq: I agree that we should maybe hold off on changing output structures until we've decided on a uniform way to do this for all functions.  E.g., an argument to get the old behavior back so an update to webchem doesn't fully break people's code.

andschar: Ok, I agree with you both @stitam and @Aariq. Will change the output back to the current state. At a later point we can discuss re-structuring the "data" output.

andschar:
changed the etox.R and the test-etox.R files as discussed. Left the improved code out-commented there to continue work in the future. Don't know why CI fails.

stitam:
It failed because I didn't approve the changes, also because there were some conflicts in `extractors.R` but I resolved them. Should be fine now.

stitam:
Closes #227

PR #230, opened 2020-03-26
Implementing PubChem PUG-View web service

stitam:
The Pubchem PUG-View web service allows the programmer to access information from the PubChem content pages (e.g. https://pubchem.ncbi.nlm.nih.gov/compound/176). The proposed function which implements this functionality is `pc_sect()` as it accesses a section of the content pages (I struggled with naming this function so I am open to suggestions). An example:

```
pc_sect(c(176, 311), "pka")
```

This returns a tibble of pKa a values. Under the hood the function uses `pc_page()` to retrieve a list of content pages and `pc_extract()` to extract information from these content pages and assemble the tibble.

Some issues I feel we should discuss:
* pc_page() uses `webchem_submit()` (now renamed to `submit_request()`) to submit POST requests in a standardised manner. This function is an attempt to standardise all http requests in webchem. Do you think this function would help? Edit: removed the function, let's discuss this as a separate issue.
* `pc_page()` and `pc_extract()` use the `data.tree` package, which is a new dependency.  It would be great not to depend on another package, but I couldn't navigate the content pages otherwise. Can you suggest an alternative?

PR task list:
- [x] Update NEWS
- [x] Check package passed


R/pubchem.R
```diff
@@ -320,3 +325,216 @@ pc_synonyms <- function(query, from = "name", choices = NULL, verbose = TRUE,
...
+#' Retrieve data from PubChem content pages
+#'
+#' When you search for an entity at \url{https://pubchem.ncbi.nlm.nih.gov/},
+#' e.g. a compound or a substance, and select the record you are interested in,
+#' you will be forwarded to a PubChem content page. When you look at a PubChem
+#' content page, you can see that chemical information is organised into
+#' sections, subsections, etc. The chemical data live at the lowest levels of
+#' these sections. Use this function to retrieve the lowest level information
+#' from PubChem content pages.
+#' @param id numeric or character; a vector of identifiers to search for.
```
andschar, line 338: can these be any identifiers? Or do they have to be PubChem specific?

R/utils.R
```diff
@@ -409,3 +424,57 @@ chooser <- function(x, choices){
...
+#' @return An object of class response, or NA.
+#' @author Tamas Stirling, \email{stirling.tamas@@gmail.com}
+#' @examples
+#' # might fail if API is not available
+#' \donttest{
+#' webchem_submit(query = 176, qurl = paste0("https://pubchem.ncbi.nlm.nih.gov/",
+#' "rest/pug_view/data/compound/176/json?heading=pka"))
+#' }
+#' @noRd
+send_request <- function(query,
```
andschar, line 449: I'm not convinced by this function. In my opinion it is more complicated to read in comparison to `httr::POST()` and the only advantage, if I see it correctly is that the sleeping time is added.
This function originated from the discussion in #241, to provide a unified user-agent, right?
Maybe and option would also be to just check that every webchem function has a `user_agent()` in `POST()` or `GET()`?

andschar: Oh, and the function handles the `verbose` output.
Maybe it would be better to have a function only for "sleeping" and "verbosing" separately from `GET()`/`POST()`/`read_html()`etc. ?

R/pubchem.R
```diff
@@ -320,3 +325,216 @@ pc_synonyms <- function(query, from = "name", choices = NULL, verbose = TRUE,
...
+#' pc_sect(c(176, 311), "density")
+#' pc_sect(2231, "depositor-supplied synonyms", "substance")
+#' pc_sect(780286, "modify date", "assay")
+#' pc_sect(9023, "Ensembl ID", "gene")
+#' pc_sect("1ZHY_A", "Sequence", "protein")
+#' pc_sect("US2013040379", "Patent Identifier Synonyms", "patent")
+#' }
+#' @export
+pc_sect <- function(id,
+                    section,
```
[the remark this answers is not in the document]
andschar: very nice function, btw :)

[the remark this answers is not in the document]
stitam: Thanks for the review @andschar! Yes, there are too many, but I don't know if this is good or bad:) This is the full list of options: https://pubchem.ncbi.nlm.nih.gov/classification/#hid=72. What makes `match.arg()` difficult is that for any query, only a subset of these will be available. Maybe I could add a message to all invalid requests that point to this url?

R/pubchem.R
stitam: They have to be PubChem specific, thanks for pointing this out.

R/utils.R
stitam: You may be right. For `POST` and `GET` requests, it is probably easier to just keep in mind that some arguents must be added, e.g. `user_agent("webchem")`. Although we might want to change the text later and then it is simpler to edit only one function. Also another issue I've recently found is that `httr` leaves connections open by default which can lead to complications. Maybe we could create a webchem specific configuration list and use it within our requests, e.g. `POST(url, body, config=webchem_request()`?

stitam: Waiting: We might want to change our waiting policy and it is simpler to edit only one function. However, this can also be done by a separate waiting function in utils (given that we consider this a consistency upgrade).

stitam: Verbosing is more difficult because we print a message before a request (searching), and after a request (found). Another thing this function does is NA handling. Maybe it is better to remove this functionality from this PR and discuss in a separate issue.

R/utils.R
```diff
@@ -409,3 +424,57 @@ chooser <- function(x, choices){
...
+#' @param qurl character; the query url.
+#' @param sleep numeric; the minimum timeout between requests.
+#' @param verbose logical; print message during processing to console?
+#' @param ... additional arguments for http request functions.
+#' @return An object of class response, or NA.
+#' @author Tamas Stirling, \email{stirling.tamas@@gmail.com}
+#' @examples
+#' # might fail if API is not available
+#' \donttest{
+#' webchem_submit(query = 176, qurl = paste0("https://pubchem.ncbi.nlm.nih.gov/",
```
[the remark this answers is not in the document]
stitam: Removed the function for now, I'll open a separate issue for it.

R/utils.R
stitam: Removed the function for now, I'll open a separate issue for it.

R/pubchem.R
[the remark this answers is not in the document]
andschar: Fair enough. I think it's ok, to leave it as is. The function is then just less strict.

R/utils.R
andschar: I like the idea of a webchem specific config object. Must have a look what can be put in such an object though.

andschar: Concerning waiting: I think waiting is very API specific. Some ask to wait 0.2 sec, others more or less. So, I think it's better to define this in each script individually.

andschar: Yeah, good idea to discuss this separately.

R/pubchem.R
```diff
@@ -320,3 +325,234 @@ pc_synonyms <- function(query, from = "name", choices = NULL, verbose = TRUE,
...
+        info <- data.frame(String = paste(y$value, collapse = " "),
+                          ReferenceNumber = y$ReferenceNumber,
+                          stringsAsFactors = FALSE)
+        return(info)
+      }
+      else{
+        string <- sapply(lownode, function(z) z$String)
+        info <- data.frame(String =  string,
+                           ReferenceNumber = y$ReferenceNumber,
+                           stringsAsFactors = FALSE)
```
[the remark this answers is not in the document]
stitam: Thanks @Aariq for your review. I didn't use `section` as column name because some section names are quite long, e.g. `pc_sect(139593369, "Evaluations of the Joint FAO/WHO Expert Committee on Food Additives - JECFA")`. On the other hand, `String` is not informative, so I renamed it to `Result`. I couldn't reproduce the messages, but using tibbles all the way is cleaner, so I changed data frames to tibbles.

R/pubchem.R
```diff
@@ -320,3 +325,234 @@ pc_synonyms <- function(query, from = "name", choices = NULL, verbose = TRUE,
...
+#' these sections. Use this function to retrieve the lowest level information
+#' from PubChem content pages.
+#' @param id numeric or character; a vector of PubChem identifiers to search
+#' for.
+#' @param section character; the section of the content page to be imported.
+#' @param domain character; the query domain. Can be one of \code{"compound"},
+#' \code{"substance"}, \code{"assay"}, \code{"gene"}, \code{"protein"} or
+#' \code{"patent"}.
+#' @param verbose logical; should a verbose output be printed on the console?
+#' @return Returns a tibble of query results.
```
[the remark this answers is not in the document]
stitam: Done.

R/ping.R
```diff
@@ -29,6 +29,24 @@ ping_pubchem <- function(...) {
...
+#' @export
+#' @examples
+#' \dontrun{
+#'  # might fail if API is not available
+#'  ping_pubchem_pw()
+#'  }
+ping_pubchem_pw <- function(...) {
+  qurl <- paste("https://pubchem.ncbi.nlm.nih.gov/rest/pug_view/data",
+               "compound/176/JSON?heading=pka", sep = "/")
+  res <- POST(qurl, user_agent("webchem"), timeout(10))
```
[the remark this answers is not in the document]
stitam: Done.

R/pubchem.R
```diff
@@ -320,3 +325,234 @@ pc_synonyms <- function(query, from = "name", choices = NULL, verbose = TRUE,
...
+#' Retrieve data from PubChem content pages
+#'
+#' When you search for an entity at \url{https://pubchem.ncbi.nlm.nih.gov/},
+#' e.g. a compound or a substance, and select the record you are interested in,
+#' you will be forwarded to a PubChem content page. When you look at a PubChem
+#' content page, you can see that chemical information is organised into
+#' sections, subsections, etc. The chemical data live at the lowest levels of
+#' these sections. Use this function to retrieve the lowest level information
+#' from PubChem content pages.
+#' @param id numeric or character; a vector of PubChem identifiers to search
```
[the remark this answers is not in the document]
stitam: This is intentional, you can query compounds, substances, assays, patents, etc. Also depending on the domain, the name of the id column in the resulting tibble will change, e.g for `pc_sect(176, "pka")` it will be `CID`, but for `pc_sect("US2013040379", "Patent Identifier Synonyms", "patent")` it will be `PatentID`.

stitam:
I updated the PR based on @Aariq's comments. The package is still not building, I think this has to do with AppVeyor being unable to check the examples because some of the webservices are down.

R/pubchem.R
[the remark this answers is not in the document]
Aariq: I like `Result` better!

Issue #228, opened 2020-03-24
build_aw_idx() should be automated


stitam:
`aw_query()` uses an index which is stored in data/aw_idx.rda and is updated through `build_aw_idx()`. It would be great if index updating could be automated. At its current state, this would involve automatically writing to the user's/contributor's hard drive which is not good practice. A better alternative would be to remove the data file from the package completely, build the index whenever aw_query() is first used in a session, and store the index in tempdir(). Other suggestions?

PR #233, opened 2020-04-03
build_aw_idx to tempdir(), remove aw_idx from package

stitam:
This PR is a solution to issue #228. `aw_query()` calls `build_aw_idx()` which builds a new index and saves it to `tempdir()` if 1) there is no index there already or 2) the user forces a new index to be built or 3) the index is older than 30 days. This automates index building and so manually updating aw_idx and storing it in the package as a data file is no longer necessary.

PR task list:
- [x] Update NEWS
- [x] Check package passed


R/alanwood.R
```diff
@@ -8,9 +8,8 @@
 #' @param  query character; search string
 #' @param type character; type of input ('cas' or 'commonname')
 #' @param verbose logical; print message during processing to console?
-#' @param idx data.frame; index to use. If NULL (default) the internal index
-#' \code{\link{aw_idx}} is used. To rebuild the index use
-#' \code{\link{build_aw_idx}}.
+#' @param force_build logical; force building a new index? See
+#' \code{build_aw_idx()} for more details.
```
Aariq, line 12: add a link to this help file.
```suggestion
#' \code{\link{build_aw_idx()}} for more details.
```

R/alanwood.R
```diff
@@ -137,42 +134,61 @@ aw_query <- function(query, type = c("commonname", "cas"), verbose = TRUE,
...
-  ln <- aw_idx$linknames
-  Encoding(ln) <- "latin1"
-  ln <- iconv(ln, from = "latin1", to = "ASCII", sub = "")
-  aw_idx$linknames <- ln
-  attr(aw_idx, "date") <- Sys.Date()
+#' @seealso \code{\link{aw_query}}
+#' @author Eduard Szoecs, \email{eduardszoecs@@gmail.com}
+#' @source \url{http://www.alanwood.net/pesticides}
+build_aw_idx <- function(verbose = TRUE, force_build = FALSE) {
+  message(msg = "build_aw_idx() will not be exported in future releases.")
```
Aariq, line 149: it's already not exported in this PR.  Is that intentional?

R/alanwood.R
```diff
@@ -137,42 +134,61 @@ aw_query <- function(query, type = c("commonname", "cas"), verbose = TRUE,
...
 #' Function to build index
 #'
+#' This function builds an index of Alan Woods Compendium of Pesticides
+#' \url{http://www.alanwood.net/pesticides} and saves it to \code{tempdir()}.
+#' This is a utility function for \code{aw_query()} and will not be exported in
+#' future releases.
 #' @import xml2
+#' @param verbose logical; print message during processing to console?
+#' @param force_build logical; force building a new index?
 #' @return a data.frame with three columns: cas, url, and name
```
Aariq, line 144: is this accurate?  I get names, links, linknames, and source.  Not sure I know what `source` represents either.

R/alanwood.R
```diff
@@ -137,42 +134,61 @@ aw_query <- function(query, type = c("commonname", "cas"), verbose = TRUE,
...
+    idx4 <- read_html("http://www.alanwood.net/pesticides/index_cn.html")
+    n <- xml_find_all(idx4, "//a")
+    names <- xml_text(n)
+    rm <- names == ""
+    names <- names[!rm]
+    links <- xml_attr(n, "href")
+    links <- links[!rm]
+    idx4 <- data.frame(names = NA, links = links, linknames = names,
+                       source = "cn", stringsAsFactors = FALSE)
+    aw_idx <- rbind(aw_idx, idx4)
```
Aariq, line 180: this line errors for me when I run `build_aw_idx()` in a fresh R session:
```
Error in rbind(aw_idx, idx4) : object 'aw_idx' not found
```

R/alanwood.R
stitam: No, thanks. Fixed.

R/alanwood.R
stitam: This was caused by the `if` condition in `build_aw_idx()` and was fixed in line 155.

R/alanwood.R
stitam: I tried to change the internals of the functions as little as possible and I didn't change the structure of the returned object. I believe `rn`stands for registry number. btw I don't see how anyone would use this function outside of `aw_query()`. List all compounds present in the database by CAS? Maybe it would be better to hard switch `build_aw_idx()` to internal and leave a note in NEWS. What do you think?

Aariq: I'd maybe just leave it as exported for now.  Someone might want to manually re-build sometime when the force_build doesn't trigger automatically.

PR #234, opened 2020-04-04
Update cs_control to query all ChemSpider data sources by default

stitam:
The ChemSpider API behind `cs_formula_csid()` uses the datasources argument in `cs_control()` to limit the scope of the search. I updated `cs_control()` function to include all possible sources. `get_csid(x, from = "formula")` now returns the same ids as you would get from a search on the website.

PR task list:
- [x] Update NEWS
- [ ] Check package passed


stitam:
This PR is related to #206.

Issue #237, opened 2020-04-15
search by CAS or formula in get_etoxid()


Aariq:
Looking at the web interface for ETOX, it looks like it should be possible to search by things other than name.  I can't find any documentation for the API (in English, at least), so I don't know how to implement this.

andschar:
Good point! I always thought this wasn't possible but now I checked again and I got it running. Can file a PR this week :)

PR #241, opened 2020-04-20
added CAS, EC, GSBL, RTECS search options to get_etoxid()

andschar:
I've added features to query the ETOX data-base by CAS, EC, GSBL, RTECS numbers in addition to the already existing option to search for names (as discussed in #237).

NB: I've added examples and tests for all the new features except for RETCS-numbers, which seems to be a proprietary format, for which I only found rather old publications. I left it there nonetheless.

PR task list:
- [x] Update NEWS
- [x] Check package passed


NEWS
```diff
@@ -3,13 +3,14 @@ webchem 0.5.0.9000
 
 NEW FEATURES
 
+* get_etoxid() now can search by CAS, EC, GSBL and RTECS numbers. Added `from = ` argument. [PR #240, added by @andschar]
```
andschar, line 6: It's PR #241. Sorry, will change.

R/etox.R
```diff
@@ -43,81 +50,93 @@ get_etoxid <- function(query,
...
   }
+  # checks
+  from <- match.arg(from)
   match <- match.arg(match)
-  foo <- function(query, match, verbose) {
+  foo <- function(query, from, match, verbose) {
     if (verbose)
       message("Searching ", query)
     baseurl <- "https://webetox.uba.de/webETOX/public/search/stoff.do"
+    if (from == 'name') {
```
andschar, line 60: this if-clause is necessary b/c the Etox-DB search endpoint has two separate fields. One for entering names, the other for number, such as (CAS, EC, etc.)

R/etox.R
```diff
@@ -43,81 +50,93 @@ get_etoxid <- function(query,
...
       d <- NA
     }
     if (length(subs) > 0) {
-      # types can be ETOX_NAME, SYNONYM
-      # the function currently uses all types, there is no filtering
-      type <- clean_char(xml_text(xml_find_all(
-        tt, "//*/table[@class = 'listForm resultList']/tr/td[2]")))
-      if (!'ETOX_NAME' %in% type) {
-        warning('No ETOX_NAME found. Returning SYNONYM.')
-      }
```
andschar, line 72: Removed this part. It would warn the user that no ETOX_NAME but only a SYNONYM was found. Imho, this is not useful, since, `get_etoxid()` is there to search for IDs and not to compare names.

R/etox.R
```diff
@@ -30,8 +33,12 @@
 #' comps <- c("Triclosan", "Glyphosate")
 #' get_etoxid(comps)
 #' get_etoxid(comps, match = "all")
+#' get_etoxid("34123-59-6", from = "cas") # Isoproturon
+#' get_etoxid("133483", from = "gsbl") # 3-Butin-1-ol
+#' get_etoxid("203-157-5", from = "ec") # Paracetamol
 #' }
 get_etoxid <- function(query,
+                       from = c("name", "cas", "ec", "gsbl", "rtecs"),
```
Aariq, line 41: The documentation should explain what these options mean (especially the last 3) either in @params or @details or at the very least point the user to a URL where they can learn.

R/etox.R
```diff
@@ -43,81 +50,93 @@ get_etoxid <- function(query,
...
+                   'stoffnummer.selection[0].type' = type,
+                   event = "Search")
+    }
     Sys.sleep(rgamma(1, shape = 15, scale = 1/10))
-    h <- POST(url = baseurl, body = list("stoffname.selection[0].name" = query,
-                                    event = "Search"))
-    # get substances and links
+    h <- POST(url = baseurl,
+              handle = httr::handle(''), # deletes cookies
+              body = body)
```
Aariq, line 77: I noticed that in wikidata.R the `POST` function uses `user_agent()` to identify the webchem package to the server.  I don't know much about `httr`, but is that something that we should be standardizing across the package?  I don't even know if it's something that makes sense for `POST` because I've only seen it in that one place.

R/etox.R
andschar: Agreed and updated.

R/etox.R
andschar: I agree with you, this should probably be standardized across the package. I reverted it to the state it was before (i.e. no `handle()`). I put it there in the first place because cookies caused some annoyances while debugging. Let's discuss this elsewhere maybe.

andschar:
don't know why travis checks are failing. Could it be because of man/nist_ri.Rd? It doesn't seem to be built.

stitam:
According to log it has to do with `cts_compinfo()`. The server seems to be down.

R/etox.R
stitam: I absolutely agree with standardizing this. An interesting link I found on the topic: https://cran.r-project.org/web/packages/httr/vignettes/api-packages.html. `httr::GET()` and `httr::POST()` requests are used by `chebi`, `chemspider`, `cir`, `etox`, `opsin`, `pubchem`, `srs`, and `wikidata` functions so they are quite common in the package. For PR #230 I created a function that aims to help us standardise these requests and related verbose outputs. I'll push it soon so we can review and discuss it?

andschar: Sounds good! Just @ us once you push your PR. Thanks for the link.

R/etox.R
stitam: If I search for "benzene", the website prints a lot of substances, some are classified as ETOX_NAME, others as SYNONYM or GSBL.  On of the responses is 1,3,5-Trimethylbenzol (https://webetox.uba.de/webETOX/public/basics/stoff.do?id=1398), classified as ETOX_NAME, CAS 108-67-8, while another is 1,3,5-Trimethylbenzen (https://webetox.uba.de/webETOX/public/basics/stoff.do?id=9008), classified as SYNONYM, CAS is again 108-67-8.  Seems like these two records refer to the same substance and the SYNONYM actually provides more information, i.e. EC and GSBL. I am not sure what the difference is between ETOX_NAME, SYNONYM, and GSBL here. Is it possible to encounter a substance where ETOX_NAME is not available but SYNONYM is, and it is a valid response?

andschar: ETOX_NAME is the name they use for the substance. SYNONYMs are synonyms for this substance. GSBL are names that are used in https://gsbl.de (for cross-referencing). When `get_etoxid(query, from = 'name')`, then the function searches in all the name fields (i.e. ETOX_NAME, SYNONYM, GSBL etc.). I don't think there is a substance without an ETOX_NAME but a SYNONYM entry.
Generally I would be rather careful with this data resource. They built it once and maintain it, but I doubt that they put a lot of energy for improvement in it.

Aariq:
> According to log it has to do with `cts_compinfo()`, the server seems to be down. I think @Aariq had the same problem in PR #232. While this is something we have to know about, maybe it shouldn't cause a build to fail. I suggest we ignore if Travis is failing because of `cts_compinfo()` but open a separate issue for the topic and discuss how to handle these situations.

It only fails on the development build of R.  It passes with the release build of R.  I can't figure out why either.

R/etox.R
```diff
@@ -43,81 +52,92 @@ get_etoxid <- function(query,
...
+  from <- match.arg(from)
   match <- match.arg(match)
-  foo <- function(query, match, verbose) {
+  foo <- function(query, from, match, verbose) {
     if (verbose)
       message("Searching ", query)
     baseurl <- "https://webetox.uba.de/webETOX/public/search/stoff.do"
+    if (from == 'name') {
+      body <- list("stoffname.selection[0].name" = query,
+                   "stoffname.selection[0].type" = "",
```
stitam, line 64: When I run the example `get_etoxid("triclosan")` it works, even when I run it multiple times. After that, when I run `get_etoxid("34123-59-6", from = "cas")`, it doesn't work, and when I subsequently try `get_etoxid("triclosan")`, it doesn't work either. The same happens the other way around. As long as I run `get_etoxid()` with `from != "name"` they all work, I can even switch between them, but as soon as I change to `from = "name"` everything stops working. When I start a fresh session, it works for a while, but then it breaks, always when switch between the `from` alternatives. Do you experience the same issue? Could it be that for some reason we are not closing the connection between requests?

andschar: My guess is, this is related to cookies. I experienced similar behaviour when debugging, but thought when running the complete function that this is not occurring anymore. Meh, it does.. This can immediately be resolved by adding `handle = handle('')` to `httr::POST()` as suggested [here](https://stackoverflow.com/questions/39979393/how-to-remove-cookies-preserved-by-httrget). What do you think?

stitam: It seems to be working, great!

andschar: Updated. Sorry, I pulled the new changes before pushing the handle() commit.

PR #242, opened 2020-04-22
get_*() consistency

Aariq:
## Pull Request

This addresses #193 and implements a consistent API and output (tibble) format for all `get_*()` functions.
It also adds a new (non-exported) utility function, `matcher` that maybe should superseed `chooser` eventually.  This should be able to handle the `match` argument of *most* functions.

PR task list:
- [ x ] Update NEWS
- [ x ] Update documentation with `devtools::document()`
- [ x ] Check package passed


tests/testthat/test-chemspider.R
```diff
@@ -51,19 +51,19 @@ test_that("get_csid()", {
...
+  c4 <- get_csid("Oxygen", order_by = "referenceCount")
+  c5 <- get_csid("Oxygen", order_by = "dataSourceCount")
+  c6 <- get_csid("Oxygen", order_by = "pubMedCount")
+  c7 <- get_csid("Oxygen", order_by = "rscCount")
+  c8 <- get_csid("Oxygen", order_direction = "ascending")
+  c9 <- get_csid("Oxygen", order_direction = "descending")
+  f <- get_csid("C47H93N2O6P", from = "formula",
+                order_by = "dataSourceCount",
+                order_direction = "descending"
+  )
```
stitam, line 66: Some of the tests are failing. The tibbles seem to be arranged in increasing csid order, even when I explicitly specify `order_direction = "descending"`, can this be a `tibble` thing? Also `get_csid("C47H93N2O6P")` in line 63 is failing. I tried to play with it to find out why, but I exceeded my monthyl allowance during the investigation:)

tests/testthat/test-wikidata.R
```diff
@@ -12,17 +12,19 @@ test_that("get_wdid returns correct results", {
...
+  expect_s3_class(o4, 'data.frame')
+
+  expect_equivalent(o1$wdid, c("Q163648", "Q57731093", NA, 'Q47512'))
+  expect_equivalent(o2$wdid[1:2], c("Q163648", "Q949424"))
 })
 
+test_that("get_wdid() handles NAs", {
+  expect_s3_class(get_wdid(NA), "data.frame")
+  expect_s3_class(get_wdid(c("Triclosan", "Glyphosate", NA)), "data.frame")
+})
```
stitam, line 27: I like that `NA` handling is tested in a separate `test_that()` this way it is easier to see if the function has already been updated or not.

R/chemspider.R
```diff
@@ -169,30 +178,44 @@ cs_control <- function(datasources = vector(),
...
+  out <-
+    purrr::map(query,
+               ~ foo(
+                 x = .x,
+                 from = from,
+                 match = match,
+                 verbose = verbose,
+                 apikey = apikey,
+                 ...
+               ))
```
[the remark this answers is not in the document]
Aariq: Mostly I just never learned the `*apply` family of functions.  But I think it's very easy to switch to `lapply` in this instance.

tests/testthat/test-chemspider.R
Aariq: I may or may not be able to figure this out since my monthly allowance is also exceeded.  I'll look into how `order_direction` is implemented and see if there's some conflict with tibble though.

R/chemspider.R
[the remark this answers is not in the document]
Aariq: Would this be the equivalent? 

```suggestion
  out <-
    lapply(
      query,
      foo,
      from = from,
      match = match,
      verbose = verbose,
      apikey = apikey,
      ...
    )
```

R/chemspider.R
```diff
@@ -169,30 +178,44 @@ cs_control <- function(datasources = vector(),
...
   names(out) <- query
-  out <- data.frame(
-    "csid" = unlist(out),
-    "query" = rep(names(out), times = sapply(out, length)),
-    row.names = NULL,
-    stringsAsFactors = FALSE
-  )
+  out <-
+    tidyr::unnest(tibble::enframe(out, name = "query", value = "csid"),
+                  cols = c("csid"))
```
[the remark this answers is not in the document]
Aariq: Again, difficult for me to test right now with my API allowance exceeded, but I think I can replace this with something like 

```
lapply(out, enframe, name = NULL, value = "csid") %>% bind_rows(.id = "query")
```

tests/testthat/test-chemspider.R
Aariq: It might work now that I've ditched `tidyr::unnest()`, but I can't test it because of exceeded API allowance.

R/chemspider.R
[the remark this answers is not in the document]
stitam: Yes, this should give the exact same output.

tests/testthat/test-chemspider.R
Aariq: It's passing all tests on my end now.  Try again?

DESCRIPTION
```diff
@@ -33,9 +33,10 @@ Imports:
     methods,
     dplyr,
     purrr,
+    tibble,
+    tidyr,
```
stitam, line 37:
```suggestion
```

According to NAMESPACE tidyr is no longer used in the package.

R/pubchem.R
```diff
@@ -33,7 +32,11 @@
 #' \url{https://www.ncbi.nlm.nih.gov/home/about/policies/},
 #' \url{https://pubchemdocs.ncbi.nlm.nih.gov/programmatic-access}.
 #' @author Eduard Szoecs, \email{eduardszoecs@@gmail.com}
+#' @import httr
 #' @importFrom purrr map2
```
stitam, line 36:
```suggestion
#' @importFrom purrr map map2
```

R/etox.R
```diff
@@ -125,12 +125,37 @@ get_etoxid <- function(query,
...
+            message("Returning NA. \n")
+          id <- NA
+          matched_sub <- NA
+        }
+        if (match == "all") {
+          if (verbose)
+            message("Returning all matches. \n")
+          id <- gsub("^.*\\?id=(.*)", "\\1", links)
+          matched_sub <- subs[sapply(id, function(x) grep(x, subs)[1])]
+        }
```
stitam, line 140: Tests fail in `test-etox.R`. The various match options seem to be messed up. `if (length(subs) == 1)` occurs at line 94 and line 112, and the latter seems to be embedded in `if (length(subs) > 1)` from line 100. Similarly `if (match == "best")` in line 147 seems to be embedded in `if (match == "best")` starting from line 126. Some other match options also occur multiple times. @Aariq, do you see the same in your version?

Aariq: that's probably because I merged the conflicts wrong.

stitam: I thought git would mark the conflict automatically. Nevertheless, it's fixed, thanks.

PR #243, opened 2020-04-23
Tests to check examples in the article

stitam:
PR task list:
- [x] Update NEWS
- [ ] (Update documentation with `devtools::document()`)
- [x] Check package passed


stitam:
Also this PR closes #204.

Aariq:
Aren't many of these tests duplicates of tests we already have?

stitam:
Thanks @Aariq, are you refering to the examples in `test-etox.R`? I think those examples are from an earlier version and should be removed. It's probably better to put all tests in a separate test file.

Aariq:
I just meant there's double coverage of functions if we add these tests. E.g. the test on line 123 just checks if `aw_query()` is working as expected—we already have a test for that in test-alanwood.R.  That's fine, but it's going to take the tests longer to run and I'm not sure I fully understand the purpose or necessity of these tests. Is it just so that people who come here from the publication find everything working exactly as expected?

stitam:
I see. My reasoning was that we have to be more careful with tests that check outcomes in the article and only change them if absolutely necessary. For this we have to know that some tests are more imporant than the others, and I figured it's best to put them into another file and disregard duplicates. But I see your point regarding testing speed. I will try to add these tests to their original test files and remove duplicates.

stitam:
I restructured the tests and tried to remove duplicates. etox and chemspider hasn't been tested yet as the servers seem to be down at the moment.

stitam:
Some functions returned a warning about closing unused connections. I think this occurred when a scraper was asked to access a page that didn't exist (e.g. intentionally invalid query). I added `on.exit(suppressWarnings(closeAllConnections()))` to scrapers to ensure connections are closed without warning.

stitam:
Also some parts of the article had to be edited, mostly because of multiple updates to `get_cid()`.

Aariq:
> Also some parts of the article had to be edited, mostly because of multiple updates to `get_cid()`.

I promise not to drastically change outputs of functions for a little while at least...  Perhaps we should focus on bug-fixing (e.g. `ci_query`) and then make a release?

R/flavornet.R
```diff
@@ -27,6 +27,7 @@
 fn_percept <- function(CAS, verbose = TRUE, ...)
 {
   foo <- function (CAS, verbose){
+    on.exit(suppressWarnings(closeAllConnections()))
```
Aariq, line 30: Thank you so much for this.  I couldn't figure out how to deal with the weird straggler open connection warnings!

tests/testthat/test-utils.R
```diff
@@ -1,12 +1,22 @@
 library(rcdk)
+
+test_that("examples in the article are unchanged", {
+  expect_false(is.inchikey("BQJCRHHNABKAKU-KBQPJGBKS-AN"))
+  expect_equal(capture_messages(is.inchikey("BQJCRHHNABKAKU-KBQPJGBKS-AN")),
+               "Hyphens not at position 15 and 26.\n")
+  expect_false(is.cas('64-17-6'))
+  expect_equal(
+    capture_messages(is.cas("64-17-6")), "Checksum is not correct! 5 vs. 6\n")
+  expect_false(is.inchikey("BQJCRHHNABKAKU-KBQPJGBKSA-5", type = "chemspider"))
```
Aariq, line 10: Does this need a `skip_if_not()` to check if chemspider is down?  I'm not super familiar with how this one works.

stitam:
> I promise not to drastically change outputs of functions for a little while at least...

I am glad the `get_*()` functions are now consistent, it was definitely worth all the trouble.

tests/testthat/test-utils.R
stitam: Thanks, I didn't notice this function accesses a web service.

stitam:
Some tests seem to fail randomly on AppVeyor or Travis. All of these tests work locally..

Aariq:
The appveyor failure seems like maybe a Windows + curl issue that's not unique to R: https://github.com/certbot/certbot/issues/7322.  The Travis one looks like it's sometimes just not connecting to CTS?  CTS has been quite slow lately.  Maybe the timeout needs to be longer for this function?

Issue #209, opened 2020-02-07
ACToR


stitam:
"EPA’s Aggregated Computational Toxicology Online Resource (ACToR) aggregates data from thousands of public sources on over 500,000 chemicals. It is searchable by chemical name and other identifiers. ACToR is also the data and web applications warehouse for EPA’s computational toxicology information which includes high-throughput screening, chemical exposure, sustainable chemistry (chemical structures and physicochemical properties) and virtual tissues data."

https://actor.epa.gov/actor/home.xhtml

andschar:
In the review process of my current paper I was pointed to this resource and I just wrote a function to query it. Expect a PR soon :)

PR #247, opened 2020-05-05
Dev actor comptox

andschar:
## Pull Request

That's the first part of the the PR to include the AcTOR data source into webchem (Issue #209). It's not yet finished (documentation etc. missing) and here for discussion.

I haven't found any non-allowances and generally the EPA has rather open policies about their data, though this is still web scraping and not an official API. Probably best to ask them.

Once we have decided on this source, I will update the PR.

PR task list:
- [ ] Update NEWS
- [ ] Check package passed

### DIFF
--- a/R/actor.R
+++ b/R/actor.R
@@ -1,0 +1,165 @@
+#' Retrieve data from AcTOR
+#'
+#' @description The function allows to search the the
+#' Aggregated Computational Toxicology Online Resource (AcTOR), which is the
+#' warehouse for different US EPA web applications: \url{https://actor.epa.gov}.
+#' Only the top most parameters (e.g. dsstox id, inchi) are queried because
+#' more advanced parameters (e.g. Hazard, Acute Toxicity, Use) only represent
+#' a loose collection of un-structure data. It is recommended to look these up
+#' manually. Entries can only be queried by a valid CASNR.
+#'
+#' @import httr xml2
+#'
+#' @param query character; search term.
+#' @param from character; type of input. Only "cas".
+#' @param verbose logical; should a verbose output be printed on the console?
+#'
+#' @references \url{https://actor.epa.gov}
+#' @author Andreas Scharmueller, \email{andschar@@protonmail.com}
+#' @return data.frame tibble
+#' @export
+#' @examples
+#' \donttest{
+#' # might fail if web site is not available
+#' actor("1071-83-6")
+#'
+#' # multiple inputs
+#' comp <- c("1071-83-6", "50-00-0")
+#' actor(comp)
+#'
+#' }
+#'
+actor <- function(query,
+                  from = "cas",
+                  verbose = TRUE) {
+  # checking
+  from <- match.arg(from)
+  foo <- function(query,
+                  from,
+                  verbose) {
+    # url
+    baseurl <- "https://actor.epa.gov/actor/chemical.xhtml"
+    prolog <- "?casrn="
+    qurl <- paste0(baseurl, prolog, query)
+    # query
+    if (verbose)
+      message("Querying: ", query)
+    site <- try(read_html(qurl))
+    if (inherits(site, "try-error")) {
+      message("Error. Returning NA.")
+      out <- data.frame(query = query,
+                        stringsAsFactors = FALSE)
+    }
+    # prepare
+    chemical_name <- trimws(xml_text(xml_nodes(site, ".chemicalNameFont")))
+    cas_dsstox <- xml_nodes(site, "#dsstoxSubstanceIdContainerId")
+    cas <- trimws(xml_text(xml_node(cas_dsstox, "#casrnId")))
+    dsstox <- trimws(xml_text(xml_child(cas_dsstox[[1]], 3))) # error prone
+    inchi <- trimws(xml_text(xml_node(site, "#inchiContainerId")))
+    inchi <- trimws(sub("InChi: InChI=", "", inchi, fixed = TRUE))
+    inchikey <- trimws(xml_text(xml_node(site, "#inchiKeyContainerId")))
+    inchikey <- trimws(sub("InChi Key:", "", inchikey))
+    formula <- trimws(xml_text(xml_node(site, "#molFormulaContainerId")))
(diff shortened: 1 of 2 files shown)


R/actor.R
```diff
@@ -0,0 +1,165 @@
...
+#' warehouse for different US EPA web applications: \url{https://actor.epa.gov}.
+#' Only the top most parameters (e.g. dsstox id, inchi) are queried because
+#' more advanced parameters (e.g. Hazard, Acute Toxicity, Use) only represent
+#' a loose collection of un-structure data. It is recommended to look these up
+#' manually. Entries can only be queried by a valid CASNR.
+#'
+#' @import httr xml2
+#'
+#' @param query character; search term.
+#' @param from character; type of input. Only "cas".
```
andschar, line 14: CAS are the only parameters to be used in the query AFAIK.

R/actor.R
```diff
@@ -0,0 +1,165 @@
...
+    cas <- trimws(xml_text(xml_node(cas_dsstox, "#casrnId")))
+    dsstox <- trimws(xml_text(xml_child(cas_dsstox[[1]], 3))) # error prone
+    inchi <- trimws(xml_text(xml_node(site, "#inchiContainerId")))
+    inchi <- trimws(sub("InChi: InChI=", "", inchi, fixed = TRUE))
+    inchikey <- trimws(xml_text(xml_node(site, "#inchiKeyContainerId")))
+    inchikey <- trimws(sub("InChi Key:", "", inchikey))
+    formula <- trimws(xml_text(xml_node(site, "#molFormulaContainerId")))
+    formula <- trimws(sub("Molecular Formula:", "", formula))
+    molecularweight <- trimws(xml_text(xml_node(site, "#molWeightContainerId")))
+    molecularweight <- trimws(sub("Molecular Weight:", "", molecularweight))
```
andschar, line 65: This function only retrieves basic parameters named here. AcTOR however, contains a lot more data, though in very unstructured formats that are imho not easily parsable (links to documents, various data formats etc.).
Yet, I think it's quite useful to retrieve nice common names and the DSSTOX-ID which is used in the Comptox apllication as an identifier.

R/actor.R
```diff
@@ -0,0 +1,165 @@
...
+#' # might fail if web site is not available
+#' actor("1071-83-6")
+#'
+#' # multiple inputs
+#' comp <- c("1071-83-6", "50-00-0")
+#' actor(comp)
+#'
+#' }
+#'
+actor <- function(query,
```
andschar, line 32: It's a bit like a `get_*()` function, but it retrieves no actual AcTOR-ID (there is none). Hence I sticked to the `get_*()` style, apart from the function name.

R/actor.R
```diff
@@ -0,0 +1,165 @@
...
+#' \donttest{
+#' # might fail if web site is not available
+#' actor_img("1071-83-6")
+#'
+#' # multiple inputs
+#' comp <- c("1071-83-6", "50-00-0")
+#' actor_img(comp)
+#' }
+#'
+actor_img = function(query,
```
andschar, line 118: I just added this functionality because it was possible. So tell me what you think, whether we need such image functions.

andschar:
Ohoh, I searched a little more and found a robots.txt here:
https://www.epa.gov/robots.txt
stating:
Disallow: ACToR
Though the robots.txt also states to aim to prevent crawling, not scraping

R/actor.R
andschar: Yes, I agree. It's probably better to name it `get_actor()`. Ok? I wouldn't name it `get_actorid()` since there is no AcTOR id.

stitam: I don't think `get_*()` style should be used, because there is no ACToR specific ID as you said. I was thinking more like `actor_query()`, or `actor_compinfo()` etc.

R/actor.R
stitam: I have just opened Issue #249 for this discussion.

R/actor.R
andschar: Ok, misunderstood you. I'm fine with `actor_query()` too.

Aariq:
> Ohoh, I searched a little more and found a robots.txt here:
> https://www.epa.gov/robots.txt
> stating:
> Disallow: ACToR
> Though the robots.txt also states to aim to prevent crawling, not scraping

The robots.txt here also disallows all user agents from scraping: https://actor.epa.gov/robots.txt

Since they've gone through the trouble of making the database available in their "download" tab, they probably don't mind web scraping, but I think it would be better to ask.

andschar:
> > Ohoh, I searched a little more and found a robots.txt here:
> > https://www.epa.gov/robots.txt
> > stating:
> > Disallow: ACToR
> > Though the robots.txt also states to aim to prevent crawling, not scraping
> 
> The robots.txt here also disallows all user agents from scraping: https://actor.epa.gov/robots.txt
> 
> Since they've gone through the trouble of making the database available in their "download" tab, they probably don't mind web scraping, but I think it would be better to ask.

I have sent them a mail. Let's just wait for the reply.

stitam:
>The robots.txt here also disallows all user agents from scraping: https://actor.epa.gov/robots.txt

Do you think this affects the ACToR web service as well? https://actorws.epa.gov/actorws/

andschar:
> > The robots.txt here also disallows all user agents from scraping: https://actor.epa.gov/robots.txt
> 
> Do you think this affects the ACToR web service as well? https://actorws.epa.gov/actorws/

Holy, I have completely not seen the AcTOR webservice! Where have you found it? Seems not too well documented^^.
I guess this makes my function obsolete and we could everything via the webservice.

I think the robots.txt doesn't have an influence on a webservice.

stitam:
It was difficult to find, I admit. I found it through this link: https://actor.epa.gov/actor/download.xhtml

andschar:
> It was difficult to find, I admit. I found it through this link: https://actor.epa.gov/actor/download.xhtml

Now that's really confusing. I have been at this site several times and always thought that there one could only download the SQL dump: _actor_2015q3.sql.gz_. Have never looked into _Details_.

I wrote them a second mail and asked them about the current state of the web service.
Can change the function afterwards to use the web service.

Issue #249, opened 2020-05-06
Images


stitam:
Some web services allow us to retrieve images of substances. There are already implementations in PR #235 and PR #247. Also ChemSpider and PubChem both offer such functionality. I thought it would be good to open a small discussion about design considerations to ensure consistency.

* function name alternatives: `*_view()`, `*_viewer()`, `_*img()`, `*_image()`. Let's choose one and stick to it. I prefer `*_img()`, as proposed by @andschar, e.g. `actor_img()`, `pc_img()`.
* arguments: `query`, `from`, `format`, `width`, `height`, `verbose`, in this order?
* outputs: 1. a list of raw character vectors. 2. exported images. Alternatively, we can add two more arguments, one to decide if we want to download, e.g. `output = c('image', 'download')` as suggested by @gjgetzinger, and another for `path`.

On first thought, I think I would prefer these functions to return a list of raw character vectors. I tried png, svg, jpeg, these can all be accessed e.g. through `httr::content(httr::GET(url), type = "image")`. Raw vectors can then be rendered easily with image processing packages like `magick`, e.g. `plot(as.raster(image_read(raw)))` and also exported in any format, but we wouldn't have to import them as dependencies, so webchem could remain lightweight.

What do you think?

andschar:
I was about to open the same issue^^

- I agree with point one, to choose one and I also prefer `img`. However, maybe it's better to turn the name around like in `get_*()` to have such a naming convention `img_actor()`. What do you think?
- Ma idea for mandatory arguments:
  - `query`
  - `from`
  - `directory` argument to allow the user to store the image at a specific place
    - default write location: `tmepdir()` b/c it's OS independant
    - or name it `locaiton`?
  - `format` (e.g. svg, png, jpeg)
  - `width` & `height`
    - usually provided by the API anyway
  - apart from that any number of arguments can follow the above ones
  - `verbose` argument at the end

andschar:
IUp to now, I haven't thought of the `output` argument. Why should we want to retrieve only the paths and not directly store the image to disk?

andschar:
There is also the [CIR](https://cactus.nci.nih.gov/blog/?p=136) service which provides this functionality.
I will include a function for it in PR #247.

stitam:
[...]

PR #250, opened 2020-05-06
Add CIR Image function

andschar:
This is a PR to add image retrieval functionality from CIR to webchem discussed in #249. It's not yet finished and here for further discussion. Hence, documentation and everything meta will follow once it is decided how to proceed with the image functions. 

PR task list:
- [ ] Update NEWS
- [ ] Check package passed


R/cir.R
```diff
@@ -163,3 +163,263 @@ cir_query <- function(identifier, representation = 'smiles', resolver = NULL,
...
+        bgcolor = "red",
+        antialising = 0,
+        atomcolor = "green",
+        bondcolor = "yellow",
+        csymbol = "all",
+        hsymbol = "all",
+        hcolor = "purple",
+        header = "My funky chemical structure..",
+        footer = "..is just so awesome!",
+        frame = 1)
```
andschar, line 414: Sorry, will remove with the next update.

R/cir.R
```diff
@@ -163,3 +163,263 @@ cir_query <- function(identifier, representation = 'smiles', resolver = NULL,
...
+#' (e.g. CAS, INCHI(KEY), SMILES etc.)
+#' @param dir character; Directory to store the image.
+#' @param format character; Format of the stored image. Can be on of TODO
+#' @param format character; Output format of the image. Can be one of "png",
+#' "gif".
+#' @param width integer; Width of the image.
+#' @param height integer; Height of the image.
+#' @param linewidth integer; Width of lines.
+#' @param symbolfontsize integer; Fontsize of atoms in the image.
+#' @param bgcolor character; E.g. transparent, white, %23AADDEE
```
Aariq, line 182: escape % 
```suggestion
#' @param bgcolor character; E.g. transparent, white, \%23AADDEE
```

R/cir.R
```diff
@@ -163,3 +163,263 @@ cir_query <- function(identifier, representation = 'smiles', resolver = NULL,
...
+#' @param header character; Should a header text be added to the image? Can be
+#' any string.
+#' @param footer character; Should a footer text be added to the image? Can be
+#' any string.
+#' @param verbose logical; Should a verbose output be printed on the console?
+#' @param frame integer; Should a frame be plotted? Can be on of NULL (default)
+#' or 1.
+#' @param ... currently not used.
+#'
+#' @return an image written to disk
```
Aariq, line 200: It might be nice if it returned the file path (silently) for integration with things like `knitr::include_graphics()`.  I might want to do something like:
```r
my_img<-cir_img("CCO")
include_graphics(my_img)
```

R/cir.R
```diff
@@ -163,3 +163,263 @@ cir_query <- function(identifier, representation = 'smiles', resolver = NULL,
     out <- unlist(out)
   return(out)
 }
+
+#' Query Chemical Identifier Resolver Images
+#'
+#' A interface to the Chemical Identifier Resolver (CIR).
+#'  (\url{http://cactus.nci.nih.gov/chemical/structure_documentation}).
+#'
+#' @param query character; Search term. Can be any common chemical identifier
```
andschar, line 172: FYI there is no `from` argument b/c CIR doesn't mind which identifier you send.

R/cir.R
andschar: `%23AADDEE` works and `\%23AADDEE` doesn't.
It's passed on as a string so I don't think we need escaping here. There is also `URLencode()` anyway to handle special characters, spaces, etc.

R/cir.R
andschar: I have updated it the function to (1) write a file to disk and (2) return a data.frame with the local path and the URL

R/cir.R
Aariq: No, the documentation is malformed without the backslash

andschar: Ah ok. Thanks!

R/cir.R
```diff
@@ -163,3 +163,244 @@ cir_query <- function(identifier, representation = 'smiles', resolver = NULL,
...
+              footer,
+              frame)
+    opts <- paste0(opts, collapse = "&")
+    opts <- paste0("?", opts)
+    # url
+    qurl <- URLencode(paste0(qurl, opts))
+    # dir
+    if (is.null(dir)) {
+      dir <- tempdir()
+    }
```
stitam, line 354: Just thinking out loud. Which would be better? `tempdir()`, `getwd()`, maybe `paste0(getwd(), "/images")`. Also would it be interesting to include a sort of "update" functionality? So that the function would either `force` download all queries and overwrite existing files, or only add those which are not already present.

andschar: hm, I always hear to not use `getwd()`, hm but why not? It's definitely not as hidden as `tempdir()`.
Concerning the update funcitonality: I think this should be left to the user. I.e. to check whether a file is present or not.

stitam: You're right, I forgot the CRAN policy. "_Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed._ "

Aariq: For .png files, why not just keep them as R objects with `png::readPNG()` if no output directory is specified?  Then the function could even plot the result with `grid::grid.raster()` if `plot=TRUE`.

PR #251, opened 2020-05-07
Add CompTox webtest_query()

andschar:
Another PR that is connected to AcTOR #247 somehow. If I understand it correctly, [AcTOR](https://actor.epa.gov/actor/home.xhtml) is the underlying data warehouse for [CompTox](https://comptox.epa.gov/dashboard/dsstoxdb/results?search=DTXSID1024122), which provides many different kinds of toxicity-relevant data on chemicals. CompTox by itself doesn't have a web service, at least to my current understanding. __But__, there is one web service for a specific feature in CompTox, namely the calculation of QSAR values named [webTEST](https://www.epa.gov/sites/production/files/2018-08/documents/webtest_users_guide.pdf). For this web service I have written the function `webtest_query()`.

Some more thoughts:
- AS @stitam showed in #247 there is also an AcTOR web service. Yet I think it doesn't provide the webTEST data.
- The whole EPA chemicals, toxicity web services ecosystem seem currently quite opaque. Probably they are also currently rebuilding stuff. I also don't see how old the AcTOR web service is.
- The script file is named `comptox.R` and the function `webtest_query()` - a bit confusing, as is there web service ecosystem currently.

PR task list:
- [x] Update NEWS
- [x] Check package passed


Aariq:
One other thought: you could call the function `comptox_webtest()`?  But I agree that it's not clear if this is part of comptox or something separate.

stitam:
Thanks @andschar, wow! So the function returns estimated values right? I was wondering. There are lots of QSAR models out there, and quite a few of them are available through webservices. Opening up webchem for QSAR prediction services is somewhat new direction for us (though EPI Suite also provided predictions, and `pc_sect()` can also access computed properties from PubChem). I think this new functionality nicely fits the purpose of the package, and I am all for it, but it might include a _lot_ of new content (https://ehp.niehs.nih.gov/doi/10.1289/EHP3264) and so I think this is something we should agree on. Do we want webchem to include various predicted properties as well? Also if yes, then we should think about how we can clearly separate predicted from measured values so there is no confusion.

stitam:
Is TEST part of CompTox or is it a separate service? https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5705535/

andschar:
> One other thought: you could call the function `comptox_webtest()`? But I agree that it's not clear if this is part of comptox or something separate.

I agree, it's better than `webtest_query()`. Or, we abbreviate _comptox_? E.g. `ctx_webtest()`? Maybe there will be more CompTox functions in webchem in the future.

andschar:
> Thanks @andschar, wow! So the function returns estimated values right? I was wondering. There are lots of QSAR models out there, and quite a few of them are available through webservices. Opening up webchem for QSAR prediction services is somewhat new direction for us (though EPI Suite also provided predictions, and `pc_sect()` can also access computed properties from PubChem). I think this new functionality nicely fits the purpose of the package, and I am all for it, but it might include a _lot_ of new content (https://ehp.niehs.nih.gov/doi/10.1289/EHP3264) and so I think this is something we should agree on. Do we want webchem to include various predicted properties as well? Also if yes, then we should think about how we can clearly separate predicted from measured values so there is no confusion.

Yes, it returns QSAR predictions. Good point. Up to now, I haven't seen so many QSAR webservice. Thanks for the paper. Isn't this something for the description? Or do you want to name the functions with some pre- or suffix  like `_pred`? Or do you even think this is something for a separate package e.g. webtox? Though, I wouldn't do that, I think. Just dumping my thoughts here^^

stitam:
I know qsardb.org collects QSARs and some of them are available through their predictor API, http://qsardb.org/guidelines/predict. I don't know if there are other services available already but I think releasing your predictive model as an API is a really good way of disseminating it so I would anticipate more to come in the future. I think we are in agreement that this is in scope, right? So this PR just needs a little update and then it can be merged. 

I like the `from`, `endpoint`, `method`, `verbose` arguments, I think it covers everything we'll need for these types of queries.

PR #256, opened 2020-05-13
Simplify README and add to https://docs.ropensci.org/webchem/

### LINKED ISSUE #254: Simplify README
Our README is getting a little out of hand IMO.  I think much or all of the code could be removed (or moved to a vignette?) and the table of services could be turned into a bulleted list.  Should have title, description, how to install, contribution statement and maybe one or two examples of functionality.  Maybe one chemical translation example and one example of retrieving properties.

Aariq:
This PR cuts a **lot** of the README and adds a "getting started" vignette as well as `_pkgdown.yml` to control how https://docs.ropensci.org/webchem/ is built. It also moves `NEWS` to `NEWS.md` so that it integrates with `pkgdown` correctly.

The new README no longer has any examples of how to use `webchem`, which I think is fine because there are plenty of examples in the help files, the webchem paper (in press?), and now in a "getting started" vignette.  However, I would be happy to add a short, somewhat "generic" example to the README if you think it's necessary.  I think in the future, more in-depth examples should be developed as vignettes (which automatically get added as "articles" in the pkgdown website).

To preview the pkgdown website locally, you can run this code after cloning this PR.

```
# Install the rOpenSci theme
remotes::install_github("ropensci/rotemplate")

# Run in your package directory to build the site:
template <- list(package = "rotemplate")
pkgdown::build_site(override = list(template = template))
```

Closes #254


README.Rmd
```diff
@@ -70,332 +69,24 @@ install.packages("webchem")
...
+Without the fantastic web services `webchem` wouldn't be here. Therefore, kudos to the web service providers and developers! Please remember to acknowledge these data resources in your work using `webchem`.
 
 
 ### Related Projects
+
+- [UniprotR](https://cran.r-project.org/web/packages/UniprotR/index.html) for retrieving information on proteins from Uniprot
+- [chemspiderapi](https://github.com/RaoulWolf/chemspiderapi) provides a more bare-bones but comprehensive interface to the ChemSpider API
+- [rcdk](https://cran.r-project.org/web/packages/rcdk/index.html) allows the user to access functionality in the 'CDK', a Java framework for chemoinformatics.
+
 If you're more familiar with Python you should check out [Matt Swains](https://github.com/mcs07) repositories: [ChemSpiPy](https://github.com/mcs07/ChemSpiPy), [PubChemPy](https://github.com/mcs07/PubChemPy) and [CirPy](https://github.com/mcs07/CIRpy) provide similar functionality as `webchem`.
```
stitam, line 90: There are quite a few related projects, but how to decide which to include here? I would prefer to include only those that have a dependency relationship with `webchem`, i.e., imports, depends, suggests, regardless of direction.

Aariq: Ok, I'll revert this for now to just include the Python projects.  We could instead include a link to the CRAN task view for chemometrics?  It has links to webchem and related projects. https://cran.r-project.org/web/views/ChemPhys.html

Aariq:
> I get the following error message when trying to build the site locally:
> 
> ```
> Error in isOpen(con) : invalid connection
> In addition: Warning message:
> Topics missing from index: 
> * webchem 
> Error: callr subprocess failed: invalid connection
> ```
> 
> Are you getting the same?

Fixed, I think.

README.Rmd
stitam: I agree with adding ChemPhys, great Task View. Can you add it so we don't have to keep it in mind? After that, I think this PR is ready to be merged.

Issue #255, opened 2020-05-11
Fix tests that are currently skipped


stitam:
Some of the tests are either skipped or commented out because they fail. Such tests can be found in:

* [ ] test-chemid.R
* [ ] test-chemspider.R
* [ ] test-cts.R

Aariq:
I'm working on #224 and a couple of these are getting fixed incidentally.

stitam:
PR #243 is giving me a lot of headache. Whenever I trigger the Appveyor/Travis tests, at least one of the tests on one of the platforms fail, and this seems to be random. Out of curiosity, I checked a skipped `cts` test as well. The server is up, and the request sometimes works, sometimes not. I think this is what's causing the test failures. Maybe ping_service() is not an ideal solution and each test should be skipped based on their own HTTP status code? In this case, each function call would have to return e.g. a vector of status codes for the vector of queries. I think this could be done with `attributes`, but it would require quite some rework on most functions. What do you think?

Aariq:
CTS seems inconsistent right now.  I tried pasting the `qurl` into my browser:
http://cts.fiehnlab.ucdavis.edu/service/compound/XEFQLINVKFYRCS-UHFFFAOYSA-N.  The first time, I got an error (code 500 I think?), but subsequent times it seems to work. Maybe instead of directly converting the URL with `fromJSON(qurl)` we can use `RETRY("GET", qurl)`?

Aariq:
As for the Appveyor one that consistently produces the following error:
```
Error in curl::curl_fetch_memory(url, handle = handle) : 
  schannel: next InitializeSecurityContext failed: SEC_E_ILLEGAL_MESSAGE (0x80090326) - This error usually occurs when a fatal SSL/TLS alert is received (e.g. handshake failed).
```

I have a suspicion this has something to do with the versions of Windows and libcurl that Appveyor uses, and I'd be satisfied with `skip_on_appveyor()` for now.  I spent about 30 min trying to google this error and didn't come up with anything that looked remotely like an easy fix.

stitam:
I see the same responses here.

I never used `httr::RETRY()` before, but it looks promising!

> I'd be satisfied with skip_on_appveyor() for now. 

Do you mean skipping _all_ tests on AppVeyor?

I'll do these edits, I am super incentivised to get  PR #243 out of the way:D

stitam:
[...]

PR #259, opened 2020-05-24
handle NAs and other miscelaneous fixes

### LINKED ISSUE #224: Functions handle NAs in input incorrectly

Aariq:
This PR closes #224 and fixes one of the skipped tests in #255.

PR task list:
- [ ] Update NEWS
- [x] Check package passed


stitam:
I think your branch is behind master. Is it possible you didn't rebase your commits?

Aariq:
> I think your branch is behind master. Is it possible you didn't rebase your commits?

I really screwed up this branch.  I'm going to try and fix it, but might just start fresh if I can't get git to work.

Aariq:
I think it's good now.

Issue #263, opened 2020-06-11
'matcher()` and `match = ` should supersede `chooser()` and `choices = `?


Aariq:
I propose deprecating `choices = ` for all the functions that use it in favor of `match = `.  I know I created `choices = ` and pushed for it, but I think I like the behavior of `match = ` better now, and it would bring some consistency to the package.
I feel a little bad for the confusion this would introduce (deprecating `first` in favor of `choices`, now deprecating `choices` in favor of `match`), but it seems worth it maybe.

Aariq:
affected functions would include:

- `cts_convert()`
- `pc_synonyms()`
- `cir_query()`

In addition, `get_etoxid()` uses the `chooser()` function internally, although it has a `match` argument.  It should probably be switched to use the `matcher()` utility function and then we can completely get rid of the `chooser()` utility.

stitam:
That functionality was kind of evolving over time. I think it's fine that we experimented with it and I agree that `match` is probably the best approach we currently have. If you want to make this more consistent, feel free to go ahead, and thanks!

Aariq:
I think I got all the instances of this.  `chooser()` is still in utils.R though, so I'll leave this open until we remove it?

Issue #257, opened 2020-05-19
Change CTS to use REST API


Aariq:
While puzzling through some problems with `cts_*()` functions (see #255, #243 ) I noticed that CTS has a REST API that is not currently implemented in the `cts_*()` functions.

Aariq:
The REST API currently seems broken, but I'm watching an issue and will take a crack at this when the API is working again. https://bitbucket.org/fiehnlab/ctsproxy/issues/24/ssl-certificate-has-expired-for-rest-api

stitam:
Could `httr::RETRY()` mitigate our difficulties with the cts functions? If yes, then this may be a simple fix, I am happy to submit it.

Aariq:
If it's an easy change, then it might be a good stopgap measure until CTS gets fixed.

stitam:
Ok, then I'll jump in on this one.

stitam:
I am closing this issue as REST API would still be nice, but CTS seems to be working fine so I'll just mark it as low priority.

Aariq:
REST API seems to be working now! Features available by REST that are not currently in `webchem::cts_*` functions:

- Convert from chemical name to InChiKey with either a "popularity" or "biological" (?) score—could be integrated with the `match = ` argument for `cts_convert()`.
- Convert InChiKey to MOL/SDF
- Count biological IDs.  I think this means 'how many hits are there for this InChiKey in various biological molecule databases?' Only works with InChiKey as input.

PR #264, opened 2020-06-11
check coverage across webchem

### LINKED ISSUE #262: standardize "from"
In all of the `get_()` functions, the argument to specify what type of identifier the query is is called `from`, but in a couple of functions, this argument is `type`.

- `ci_query(query, type = "inchikey")`
- `aw_query(query, type = "cas")`

A couple of functions don't have `query` and `from` because they only accept one identifier type in queries.  Maybe these are OK as is?

- `fn_percept(CAS)`
- `cts_compinfo(inchikey)`

In addition, the possible values that `from` accepts are mostly standardized, except for:

- `aw_query`, which uses `"commonname"` instead of `"name"`
- `get_chebiid()` which is more complicated:

>character; type of input, can be one of 'ALL', 'CHEBI ID', 'CHEBI NAME', 'DEFINITION', 'ALL NAMES', 'IUPAC NAME', 'CITATIONS', 'REGISTRY NUMBERS', 'MANUAL XREFS', 'AUTOMATIC XREFS', 'FORMULA', 'MASS', 'MONOISOTOPIC MASS', 'CHARGE', 'INCHI/INCHI KEY', 'SMILES', 'SPECIES'.

It might be nice if this wasn't case-sensitive and if both "inchi" and "inchikey" were options rather than "INCHI/INCHI KEY" and maybe if "name" was a synonym for "ALL NAMES" to better standardize across the package.

Aariq:
This adds two functions that integrate multiple data sources in webchem.  `autotranslate()` is a wrapper that you can give a query of any type and the name of a webchem function and it will attempt to translate the query into an identifier the function accepts.

`check_coverage()` runs a `get_*()` or `*_query()` from most of the data sources in webchem and returns a tibble indicating which compounds got hits in each data source. 

Currently it also has the option to plot the results in a sort of table/heatmap, but the plot is made with `ggplot2`.  **I need help translating that code to base R**, because it doesn't make sense to add ggplot2 as a dependency at this time.

example

## minor changes

To make all this happen, I fixed some bugs and made the `from=` argument more consistent across the package. (closes #262 and #263)

PR task list:
- [x] Update NEWS
- [ ] Check package passed


Aariq:
I should also note that these functions rely on CTS, which currently just sometimes fail for no apparent reason.  Maybe switching to the REST API will help (#257)?

stitam:
Also sorry for introducing conflicts in your PR by merging another one..

Aariq:
As for the name, I'm not tied to `check_coverage()`.  I like `lookup()` or perhaps `has_entry()`?

>Do you imagine this function to be exported on the long term? 

I wrote this intending it to be exported and used directly by a user with a list of compounds wanting to get some common bit of data (e.g. molecular weight), but without an idea of which of several data sources to use.  But I'm actually not sure how useful this is, which is why I asked about it on Slack.  Perhaps this is just a rough sketch of some future function that integrates across data sources?

Aariq:
How about using the `plot.matrix` package instead of `ggplot2` as a compromise?  It is a much lighter weight, single purpose package, and I could add it as a Suggests rather than a Depends and prompt the user to install only if they want to plot the results.

Aariq:
The tests are failing because CTS just fails randomly sometimes.  Not sure if I should include some kind of workaround for this in the `autotranslate()` function or just ignore the failing tests and hope that CTS gets fixed sometime?

Aariq:
I went ahead and addressed #263 since it seemed in the spirit of the PR.

R/chemspider.R
```diff
@@ -191,7 +191,7 @@ get_csid <- function(query,
   match <- match.arg(match)
 
   foo <- function(x, from, match, verbose, apikey, ...) {
-    if (is.na(x)) return(NA)
+    if (is.na(x)) return(as.integer(NA))
```
stitam, line 194: `NA_integer_`?

R/chemspider.R
```diff
@@ -200,8 +200,10 @@ get_csid <- function(query,
                   inchi = cs_inchi_csid(x, apikey = apikey),
                   inchikey = cs_inchikey_csid(x, apikey = apikey),
                   smiles = cs_smiles_csid(x, apikey = apikey))
-    res <- matcher(res, query = x, match = match, verbose = verbose)
-    if (length(res) == 0) res <- NA
+    if(length(res) > 1) {
+      res <- matcher(res, query = x, match = match, verbose = verbose)
+    }
+    if (length(res) == 0) res <- as.integer(NA)
```
stitam, line 206: `NA_integer_`?

R/cts.R
```diff
@@ -89,14 +98,32 @@ cts_compinfo <- function(inchikey, verbose = TRUE){
...
-    stop('"first" is deprecated.  Use "choices = 1" instead.')
+cts_convert <- function(query,
+                        from,
+                        to,
+                        match = c("all", "first", "ask", "na"),
+                        verbose = TRUE,
+                        choices = NULL,
+                        ...){
+  if(!missing("choices"))
+    stop('"choices" is deprecated.  Use "match" instead.')
```
stitam, line 109: Previously deprecated arguments resulted in a warning message but this results in an error. Is there a reason for this?

Aariq: `choices = 10`, for example, doesn't have an exactly equivalent value for `match`. But now that you mention it, I could actually make this a message.
```suggestion
   if(!missing("choices")) {
    if (is.null(choices)) {
      message('"choices" is deprecated.  Using match = "all" instead.')
      match <- "all"
    } else if(choices == 1) {
      message('"choices" is deprecated.  Using match= "first" instead.')
      match <- "first"
    } else if ((is.numeric(choices) & choices > 1) | choices == "all") {
      message('"choices" is deprecated.  Using match = "ask" instead.')
      match <- "ask"
    } else {
      message('"choices" is deprecated.  Using match = "all" instead.')
      match <- "all"
    }
  }
```

R/alanwood.R
```diff
@@ -24,30 +26,42 @@
...
-                     force_build = FALSE) {
-  aw_idx <- build_aw_idx(verbose = FALSE, force_build)
-  foo <- function(query, type = c("commonname", "cas"), verbose) {
+
+aw_query <- function(query, from = c("name", "cas"), verbose = TRUE,
+                     force_build = FALSE, type, ...) {
+  if (!missing(type)) {
+    warning('"type" is deprecated. Please use "from" instead. ')
+    from <- type
+  }
```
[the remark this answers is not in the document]
Aariq: In this case, probably a message since it's just a change in argument.

R/chebi.R
```diff
@@ -9,24 +9,21 @@
 #' @importFrom stats setNames
 #'
 #' @param query character; search term.
-#' @param from character; type of input, can be one of 'ALL', 'CHEBI ID',
-#' 'CHEBI NAME', 'DEFINITION', 'ALL NAMES', 'IUPAC NAME', 'CITATIONS',
-#' 'REGISTRY NUMBERS', 'MANUAL XREFS', 'AUTOMATIC XREFS', 'FORMULA', 'MASS',
-#' 'MONOISOTOPIC MASS', 'CHARGE', 'INCHI/INCHI KEY', 'SMILES', 'SPECIES'.
+#' @param from character; type of input.  \code{"all"} searches all types and
+#'   \code{"name"} searches all names.
```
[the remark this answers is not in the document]
Aariq: I think I did, because now they are listed in the function call in the documentation, but I'm going to put them back so it's more clear.

R/chebi.R
```diff
@@ -9,24 +9,21 @@
...
 #' @param match character; How should multiple hits be handled?,
 #' \code{"all"} all matches are returned,
 #' \code{"best"} the best matching (by the ChEBI searchscore) is returned,
 #' \code{"ask"} enters an interactive mode and the user is asked for input,
 #' \code{"na"} returns NA if multiple hits are found.
 #' @param max_res integer; maximum number of results to be retrieved from the
 #' web service
-#' @param stars character; type of input can be one of 'ALL', 'TWO ONLY',
-#' 'THREE ONLY'.
+#' @param stars character;
```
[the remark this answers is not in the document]
Aariq: I'll replace it with a description of what the stars mean since now the acceptable inputs are in the function call.

R/chebi.R
```diff
@@ -138,29 +140,31 @@ get_chebiid <- function(query,
...
                 from = from,
+                match = match,
                 max_res = max_res,
                 stars = stars,
                 verbose = verbose)
   out <- setNames(out, query)
-  out <- as_tibble(bind_rows(out))
-  return(out)
+  out <- bind_rows(out)
+  return(dplyr::select(out, query, chebiid, everything()))
```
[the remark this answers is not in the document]
Aariq: I also get this.  The simple fix is to add an `as_tibble()`, but the "better" fix is probably to re-write this to use the `matcher()` utility instead.

NEWS.md
```diff
@@ -1,8 +1,16 @@
 # webchem 1.0.0.9001
 
-## NEW FEATURES
```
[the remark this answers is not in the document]
Aariq: fixed

R/chebi.R
```diff
@@ -67,24 +64,30 @@
...
-    from <- match.arg(from, from_all)
-    stars_all <- c('ALL', 'TWO ONLY', 'THREE ONLY')
-    stars <- match.arg(stars, stars_all)
+  from <- toupper(match.arg(from))
+  if (from == "NAME") {
+    from <- "ALL NAMES"
+  }
+  if (from == "inchi" | from == "inchikey") {
+    from <- "INCHI/INCHI KEY"
+  }
```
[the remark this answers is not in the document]
Aariq: I tried the example in the help file on the release version and it gives NA for the chebiid.  So I don't think it's anything I did.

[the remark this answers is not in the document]
stitam: Ok, then let's fix it some other time.

R/pan.R
```diff
@@ -67,8 +68,9 @@
 #'  # extract Hydrolysis Half-life (Avg, Days)
 #'  sapply(out, function(y) y$`Hydrolysis Half-life (Avg, Days)`)
 #' }
-pan_query <- function(query, match = c('best', 'all', 'first'), verbose = TRUE, ...){
+pan_query <- function(query, from = c("name", "cas"), match = c('best', 'all', 'first'), verbose = TRUE, ...){
```
stitam, line 71: One of the examples seem to be broken, `out <- pan_query(c('Triclosan', 'Aspirin'), 'best')`.

tests/testthat/test-pubchem.R
```diff
@@ -87,10 +87,10 @@ test_that("pc_synonyms", {
   skip_on_cran()
   skip_if_not(up, "PubChem service is down")
   expect_equivalent(pc_synonyms(NA), NA)
-  expect_equal(pc_synonyms("Triclosan")[[1]][1], "5564")
+  expect_equal(pc_synonyms("Acetyl Salicylic Acid")[[1]][1], "aspirin")
```
stitam, line 90: It has nothing to do with your PR, but one of the PubChem tests is broken. For `c <- pc_sect(780286, "modify date", "assay")` a new entry has been added and so `expect_equal(c$Result, c("2014-05-03", "2018-09-28"))` breaks. Could you fix this within this PR?

R/etox.R
```diff
@@ -93,61 +96,24 @@ get_etoxid <- function(query,
...
-        dd <- adist(query, msubs) / nchar(msubs)
-        id <- gsub("^.*\\?id=(.*)", "\\1", links[which.min(dd)])
-        matched_sub <- subs[which.min(dd)]
-      }
-      if (match == "ask") {
-        matched_sub <- chooser(subs, "all")
-        id <- gsub("^.*\\?id=(.*)", "\\1", links[which(subs == matched_sub)])
-      }
+
+      out <- matcher(id, query = query, result = subs_names, match = match)
```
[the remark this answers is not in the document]
Aariq: Yeah, I think most of the time going forward `matcher()` is all people will need to implement the `match` argument.  At least that's my hope.

R/pan.R
Aariq: You know, I really should remember to use `devtools::run_examples(test = TRUE, run = TRUE)`.  This might be a good reminder to add to the PR github template.

Aariq: I changed the example compounds because in addition to not running, both of these returned NA for Hydrolysis half-life.

R/integration.R
```diff
@@ -0,0 +1,118 @@
...
+#'
+#' @return returns results from \code{.f}
+#' @importFrom rlang as_function fn_fmls
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' autotranslate("XDDAORKBJWWYJS-UHFFFAOYSA-N", from = "inchikey", .f = "get_etoxid")
+#' }
+autotranslate <- function(query, from, .f, .verbose = TRUE, ...) {
```
stitam, line 20: I find the function name a bit too long for an exported function. Do you imagine this function to be used for ID conversions only? If yes, then maybe we could baldly name it `convert()`.

Do you think there would be value in updating this function later to a form where `.f` is replaced by `to`? e.g. we call `autotranslate("XDDAORKBJWWYJS-UHFFFAOYSA-N", from = "inchikey", to = "etoxid")` and the function figures out that it will need `cts_convert()` and `get_etoxid()` to perform the conversion.

R/integration.R
```diff
@@ -0,0 +1,118 @@
...
+  if (!from %in% pos_froms) {
+    pos_froms <- pos_froms[pos_froms != "name"] #cts name conversion broken
+    new_from <- pos_froms[which(pos_froms %in% cts_to())[1]]
+    if(.verbose){
+      message(
+        paste0(.f, " doesn't accept ", from, ".\n", "Attempting to translate to ", new_from, " with CTS. ")
+      )
+    }
+    new_query <- cts_convert(query, from = from, to = new_from, match = "first")
+    #would like to try a again if cts fails the first time (as it often does).
```
stitam, line 33: Good idea! Maybe `cts_convert()` should use `httr::RETRY()`?

R/integration.R
```diff
@@ -0,0 +1,118 @@
...
+#' @param .verbose logical; print a message when translating query?
+#' @param ... other arguments passed to the function specified with \code{.f}
+#'
+#' @return returns results from \code{.f}
+#' @importFrom rlang as_function fn_fmls
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' autotranslate("XDDAORKBJWWYJS-UHFFFAOYSA-N", from = "inchikey", .f = "get_etoxid")
```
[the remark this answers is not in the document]
Aariq: I see what you mean, but I think this is only possible if that the output of the function supplied to `.f` is a data frame or tibble. In the case of a tibble it would probably be better to do something like `orig_query | query | match | etoxid`.  Another idea is to make the messages more explicit when `verbose = TRUE`.  Something like 
```
get_cisd doesn't accept cas. Attempting to translate to inchikey with CTS.  
"50-29-3" -> "YVGGHNCTFXOJCH-UHFFFAOYSA-N"
"52-68-6" -> "NFACJZMKEDPNKN-UHFFFAOYSA-N"
```

R/integration.R
Aariq: So, under the hood this function converts, but the result is not necessarily just an ID conversion.  I can use, for example, `aw_query` for the `.f` argument and get a whole list of chemical properties as a result.  I'm open to name changes, but `convert` doesn't seem quite right. Would something like `with_translation()` be better?  Also, this reminded me that I should make a note that translating on the fly isn't fool-proof since only the first hit from CTS is used.

Aariq: or even `with_cts()`?

R/integration.R
Aariq: yes, maybe, but I think that should be a separate PR.

R/integration.R
```diff
@@ -0,0 +1,140 @@
...
+  sources <- match.arg(sources, several.ok = TRUE)
+  sources <- sapply(sources, switch,
+                    "etox" = "get_etoxid",
+                    "pc" = "get_cid",
+                    "chebi" = "get_chebiid",
+                    "cs" = "get_csid",
+                    "aw" = "aw_query",
+                    "fn" = "fn_percept",
+                    "pan" = "pan_query",
+                    "srs" = "srs_query")
```
stitam, line 87: Some of the potential sources are missing (e.g. nist), is there a reason for this?

R/integration.R
```diff
@@ -0,0 +1,140 @@
...
+
+  foo <- function(.f, query, from) {
+    # if a function errors (e.g. API is down) then return NA
+    x <-
+      try(autotranslate(
+        query = query,
+        from = from,
+        .f = .f,
+        match = "first"
+      ))
```
stitam, line 97: My argument against `autotranslate()` is 1.exporting, 2. new dependency, but I see the benefit of the function here. How about keeping `autotranslate()` so we don't have to rethink this otherwise great function, but not exporting it? Later we can think about whether or not we can find a solution where we don't depend on `rlang`.

Aariq: agreed that `autotranslate()` doesn't need to be exported.  I might rename it `with_cts()` if that's ok with you.  `dplyr` already imports `rlang` so I think `rlang` is already a dependency for `webchem`, indirectly.  And it's a super useful package.

R/integration.R
Aariq: NIST is missing because of the complication of the type of retention index.  Like, a chemical might have non-polar, isothermal RIs but not polar gradient RIs.  When there is a more generic NIST function, we can add it.

R/integration.R
```diff
@@ -0,0 +1,140 @@
...
+#' @export
+#' @import dplyr
+#' @examples
+#' \dontrun{
+#' has_entry("hexane", from = "name")
+#' }
+has_entry <- function(query, from,
+                           sources = c("etox", "pc", "chebi", "cs",
+                                       "aw", "fn", "pan", "srs"),
+                           plot = FALSE) {
```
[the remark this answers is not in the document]
Aariq: I'm going to do some crowd sourcing and see if I can't come up with something better.

R/integration.R
stitam: If `ȧutotranslate()` is not exported, then I don't mind the name being a bit long, but `with_cts()` is also quite descriptive, so whichever you prefer.

R/integration.R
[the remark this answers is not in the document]
Aariq: `find_db()` maybe?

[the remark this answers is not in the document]
stitam: Perfect!

Issue #269, opened 2020-07-07
When hashtag # present in SMILES, error type report as EntityRef: expecting ';' [23]


zhssakura:
Hello, 

Thanks for creating webchem. I found that when a hashtag symbol present in the SMILES, the following error type will occur:

```r
comp1 = "CC(C)CC(C)(O)C#CC(C)(O)CC(C)C"

cir_query(comp1, representation = 'smiles', resolver = NULL, first = TRUE, choices = NULL, verbose = TRUE)
#> Error in cir_query(comp1, representation = "smiles", resolver = NULL, : could not find function "cir_query"
```

Any suggestion would be helpful. Thank you.

Best wishes,

Shan

stitam:
Thanks @zhssakura for raising this issue. The error occurs because of lack of url encoding, thanks for bringing our attention to this, we'll soon fix the function. In the meantime, if you substitute `#` in your queries with `%23` it should work.

```
comp2 <- "CC(C)CC(C)(O)C%23CC(C)(O)CC(C)C"
cir_query(comp2, representation = "inchikey")
$`CC(C)CC(C)(O)C%23CC(C)(O)CC(C)C`
[1] "InChIKey=LXOFYPKXCSULTL-UHFFFAOYNA-N"
```

zhssakura:
Thanks @stitam,
I would try it as you suggested.
Best,
Shan

Aariq:
closed with PR #270.  Install the development version, @zhssakura, for a version without this bug.  Thanks again for pointing this out.

PR #270, opened 2020-07-10
Handling special characters in SMILES strings

stitam:
Related to issue #269.

In SMILES strings "#" represents a triple bond. Sometimes "#" has to be encoded as "%23". `URLencode()` does not encode this by default, so I updated all instances of the function to `URLencode(query, reserved = TRUE)`. It's important to note that it's the query that has to be encoded like this, not the full url. 

I also added tests to all functions that can query SMILES strings to check for their ability to handle special characters. 

PR task list:
- [x] Update NEWS
- [x] Check package passed


Aariq:
URL encoding queries might be a good tip to add to the contributor guidelines

Aariq:
Everything looks good to me and your explanation in contributing.md definitely helped me understand.

PR #272, opened 2020-07-14
added cir_img() function

andschar:
## Pull Request

adding the `cir_img()` function. Replacing the closed, erroneous PR #250. Alreasy uses `RETRY()` as discussed in PR #271 .

PR task list:
- [x] Update NEWS
- [not appropriate for images?] Add tests (if appropriate)
- [x] Check package passed


R/cir.R
```diff
@@ -167,3 +167,235 @@ cir_query <- function(identifier, representation = "smiles",
...
+              atomcolor,
+              bondcolor,
+              csymbol,
+              hsymbol,
+              hcolor,
+              header,
+              footer,
+              frame)
+    opts <- paste0(opts, collapse = "&")
+    opts <- paste0("?", opts)
```
Aariq, line 355: This is maybe a little more concise and I think gives exactly the same result without the "if" statements.
```suggestion
    opts <-
  c(
    format = format,
    width = width,
    height = height,
    linewidth = linewidth,
    symbolfontsize = symbolfontsize,
    bgcolor = bgcolor,
    antialiasing = as.numeric(antialiasing),
    atomcolor = atomcolor,
    bondcolor = bondcolor,
    csymbol = csymbol,
    hsymbol = hsymbol,
    hcolor = hcolor,
    header = header,
    footer = footer,
    frame = frame
  )

opts <- paste0(names(opts), "=", opts)
opts <- paste0(opts, collapse = "&")
opts <- paste0("?", opts)
```

R/cir.R
```diff
@@ -167,3 +167,235 @@ cir_query <- function(identifier, representation = "smiles",
...
+    # url
+    qurl <- URLencode(paste0(qurl, opts))
+    # query
+    if (verbose)
+      message("Querying: ", query, "\n", qurl)
+    Sys.sleep(1.5)
+    path <- file.path(dir, paste0(query, ".", sub('format=', '', format)))
+    message("Image saved under: ", path)
```
Aariq, line 363: Seems like this line should go *after* the image is successfully saved.  Where it is now, it will trigger even if the server is down, which could be misleading.

R/cir.R
```diff
@@ -167,3 +167,235 @@ cir_query <- function(identifier, representation = "smiles",
...
+    # return image
+    h <- try(
+      RETRY("GET",
+            qurl,
+            write_disk(path, overwrite = TRUE))
+    )
+    if (inherits(h, "try-error")) {
+      warning("Problem with web service encountered... Returning NA.")
+      return(data.frame(query = query, stringsAsFactors = FALSE))
```
Aariq, line 372: @stitam, is this consistent with how you're using `RETRY` in #271 ?

R/cir.R
```diff
@@ -167,3 +167,235 @@ cir_query <- function(identifier, representation = "smiles",
...
+    } else {
+      # return paths data.frame
+      data.frame(query = query,
+                 path = path,
+                 url = qurl,
+                 stringsAsFactors = FALSE)
+    }
+  }
```
Aariq, line 380: I think we might have decided that image functions don't need to return anything in #249.

R/cir.R
```diff
@@ -167,3 +167,235 @@ cir_query <- function(identifier, representation = "smiles",
...
+    if (verbose)
+      message("Querying: ", query, "\n", qurl)
+    Sys.sleep(1.5)
+    path <- file.path(dir, paste0(query, ".", sub('format=', '', format)))
+    message("Image saved under: ", path)
+    # return image
+    h <- try(
+      RETRY("GET",
+            qurl,
+            write_disk(path, overwrite = TRUE))
```
Aariq, line 368: I don't know how to fix this, but if a query fails, it seems like this function still downloads an "empty" image file.  For example:
```
cir_img("asdafsdfadf", dir = here::here())
```
produces a file, "asdafsdfadf.png", but when I try to open it, my OS says its damaged or unrecognized.

Can the `write_disk` be done in a separate step, only after `RETRY` succeeds?

Aariq:
> * [not appropriate for images?] Add tests (if appropriate)

I think it might be good to write some tests to check that error messages work as expected and file downloads happen or don't happen when expected.  You can use tempdir() to download images inside tests.  I don't think it's necessary to check that the correct image is downloaded.

R/cir.R
andschar: you are right, NULLs are dropped by default in `c()`

R/cir.R
```diff
@@ -137,7 +137,7 @@ cir_query <- function(identifier, representation = "smiles",
       if (verbose)
         message(qurl)
       Sys.sleep(1.5)
-      h <- try(GET(qurl, timeout(5)))
+      h <- try(RETRY("GET", qurl))
       if (inherits(h, "try-error")) {
         warning('Problem with web service encountered... Returning NA.')
         return(NA)
```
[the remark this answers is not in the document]
stitam: I am currently updating functions to `httr::RETRY("GET", qurl, quiet = TRUE, terminate_on = 404)`. This repeats maximum three times by default unless it succeeds, it will not print verbose messages e.g. when trying again, and it will also terminate on HTTP status code is 404 (why slow down the function when the query is "not found"). I don't think it is good practice to return an error because this would make the functions more difficult to vectorise, but let's discuss if you disagree. I think it is enough to return a HTTP message when `verbose = TRUE` e.g. with `httr::message_for_status()`.

R/cir.R
stitam: I am updating verbose messages in PR #271 and there I am using

```
if (verbose) message(paste0("Querying ", query, ". "), appendLF = FALSE)
```
last thing before the query, where `appendLF = FALSE` forces the next message to be put in the same line. Then right after the query I am using 

```
if (verbose) message(httr::message_for_status(res))
```

where res is the `RETRY()` output. This will print an informative HTTP response if the server is down.

R/cir.R
stitam: I'd rather not use try to avoid errors and warning messages.

`RETRY()` returns nice HTTP status codes which can be used to handle failures.

```
if (verbose) message(paste0("Querying ", query, ". "), appendLF = FALSE)
h <- httr::RETRY("GET", qurl,  write_disk(path, overwrite = TRUE)) # I didn't know this also works, thanks @andschar 
if (verbose) message(httr::message_for_status(res))
```

R/cir.R
stitam: Yes, I have recently added a few notes on images to [CONTRIBUTING.md](https://github.com/ropensci/webchem/blob/master/CONTRIBUTING.md).

R/cir.R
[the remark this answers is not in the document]
andschar: Thanks for the comment. I agree to not return an error.

R/cir.R
andschar: Thanks for introducing me to `RETRY()`, sounds good to me!

R/cir.R
andschar: `write_disk()` also works in `RETRY()`

R/cir.R
andschar: ok, wasn't aware anymore!

R/cir.R
andschar: I just tried that and apparently it's not possible. It has to be inside GET, RETRY or similar. Maybe we could raise an issue at the `httr` GitHub site? As mentioned above, one way would be to delete the file in an if-clause.

Aariq: Yes, this does seem kind of like a bug that `write_disk()` does something even if the `GET` fails. I suppose deleting the file if there is an html status error is a fine workaround.

andschar:
> > * [not appropriate for images?] Add tests (if appropriate)
> 
> I think it might be good to write some tests to check that error messages work as expected and file downloads happen or don't happen when expected. You can use tempdir() to download images inside tests. I don't think it's necessary to check that the correct image is downloaded.

I have added some tests to check the funciton.

andschar:
@Aariq In the update I changed the function back to if statements for all the options because some option strings need extra quotation marks, which can't be easily done with your suggestion.

R/cir.R
andschar: Hm, an issue [here](https://github.com/r-lib/httr/issues/553) brought up the same thing.

Aariq: Ah, well, it looks like this is the official workaround—check if there was an error and remove the file after the fact.
```
if (http_error(h) && file.exists(path)) {
  file.remove(path)
}
```

andschar:
to vectorise the function I now use a for loop instead of `lapply()`. This might sound "un-Rish" at first, however I think in this case it's much better because a for loop doesn't return anything. Hence, if I run `a = cir_img('CCO', tempdir())`, a will be `NULL`. If I were to use `lapply()` I'd receive some sort of list, depending on whether all results return errors or not. Also, to avoid printing something to the console, I'd have to wrap the whole function into `invisible()`.

Aariq:
One more change and I think it'll probably be good to go.  It currently has the issue where it mistakes `NA` for sodium.  
```
>cir_img(NA, here())
Querying NA. OK (HTTP 200). Image saved under: /Users/scottericr/Documents/webchem/NA.png
```
NA

R/cir.R
stitam: @Aariq suggested in Issue #249 to import images to tables through `knitr::kable()` and `pander::pandoc.image.return()`. I think this and other vectorised post processing workflows would work better if the directory contained a path for all queries not only those that were successful, so I would leave the empty images there. What do you think?

R/cir.R
```diff
@@ -167,3 +167,233 @@ cir_query <- function(identifier, representation = "smiles",
...
+        atomcolor = atomcolor,
+        bondcolor = bondcolor,
+        csymbol = csymbol,
+        hsymbol = hsymbol,
+        hcolor = hcolor,
+        header = header,
+        footer = footer,
+        frame = frame,
+        verbose = verbose)
+  }
```
stitam, line 398: How about `purrr::walk()` instead of the `for` cycle?

tests/testthat/test-cir.R
```diff
@@ -31,3 +31,15 @@ test_that("cir_query() handles special characters in SMILES", {
...
+  skip_on_cran()
+  skip_if_not(up, "CIR server is down")
+
+  expect_true(is.null(cir_img('CCO', tempdir())))
+  fl <- file.path(tempdir(), 'CCO.png')
+  expect_true(file.exists(fl))
+  fl2 <- file.path(tempdir(), 'abcdefghijk.png')
+  cir_img('abcdefghijk', tempdir())
+  expect_true(!file.exists(fl2))
+})
```
stitam, line 45: Just thinking out loud, maybe file size could be tested with `file.info(path)$size`?

R/cir.R
```diff
@@ -167,3 +167,233 @@ cir_query <- function(identifier, representation = "smiles",
...
+#'         hcolor = "purple",
+#'         header = "My funky chemical structure..",
+#'         footer = "..is just so awesome!",
+#'         frame = 1,
+#'         verbose = TRUE)
+#'}
+#' @export
+#'
+cir_img <- function(query,
+                    dir = NULL,
```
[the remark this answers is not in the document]
andschar: Agreed, will change.

R/cir.R
andschar: Hm, this would be an option, however I would generally avoid another dependency in my code (I know it's already in webchem). I think speed is also not an issue here, since it is dominated by data transmission anyway, so I would rather use `for`. If you prefer `purr::walk` for consistency with other functions, it's also ok for me..

andschar:
> One more change and I think it'll probably be good to go. It currently has the issue where it mistakes `NA` for sodium.
> 
> ```
> >cir_img(NA, here())
> Querying NA. OK (HTTP 200). Image saved under: /Users/scottericr/Documents/webchem/NA.png
> ```
> 
> NA

Good point. It's also a problem if somebody would call the function like this `cir_img('', tempdir())`, because an empty image would be returned. My solution would be the following. Check NAs and empty strings at the start of the function. What do you think? Are there better options?

```r
vec = c(NA, '', 'Triclosan')
if (anyNA(vec) || any(vec == '')) {
  stop('NA or empty strings provided.')
}
```

tests/testthat/test-cir.R
andschar: Hm, not sure whether file size, e.g. for a .png keeps constant over time?

Aariq: I think testing whether the file exists is probably sufficient.  You could also check if it is larger than 0Kb, but that would probably just be testing the CIR API more than anything.

stitam: Thanks for the feedbacks, I'm okay with testing only whether the file exists.

R/cir.R
stitam: You're right, the rate limiting step is probably data transmission, so even though `purrr` is already in webchem, if you want to stay with `for`, it should be fine.

R/cir.R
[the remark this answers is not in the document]
stitam: If you want, you can leave it like this @andschar and I'll update the API request and the verbose messages within PR #271.

R/cir.R
```diff
@@ -167,3 +167,234 @@ cir_query <- function(identifier, representation = "smiles",
...
+                    hcolor = NULL,
+                    header = NULL,
+                    footer = NULL,
+                    frame = NULL,
+                    verbose = TRUE,
+                    ...) {
+  # check
+  if (anyNA(query) || any(query == '')) {
+    stop('NA or empty string provided.')
+  }
```
stitam, line 284: With this solution `cir_img("CCO", NA)` will print an error. I think it would be better if the function downloaded the image for the first query and then just printed a verbose message for the invalid query. I think anything that stops webchem functions and prints errors should be avoided to facilitate vectorisation. If you put `NA` handling within `foo` it should be fine.

stitam (changes requested):

stitam:
Checks fail because Pubchem is a living database so tests often become outdated. Let's not worry about those tests here.

R/cir.R
andschar: I made some changes to the code that should fix this.
(1) I added an if clause before foo() to check if the dir exists and is not NA. This should be done before foo() b/c all images will be saved therein.
(2) Within foo(), I now check in the beginning whether the query string is NA or empty. If so, I send a message to the console and return NULL. Hence, the string is not queried.

stitam (approved):

Issue #277, opened 2020-08-19
plot from find_db() crops compound names


Aariq:
Someone better versed in base R plotting might be able to figure out how to get the plot margins to adjust according to the size of the axis tick labels.  I couldn't figure it out, so long compound names are cropped.

```r
library(webchem)
data("jagst")
subs <- unique(jagst$substance)
find_db(subs[1:3], from = "name", sources = c("etox", "pan", "srs"), plot = TRUE)
```
![image](https://user-images.githubusercontent.com/25404783/90684443-33007200-e236-11ea-8714-182cca0c19bb.png)

stitam:
What do you think of this?

```
leftmargin <- 0.5*max(nchar(query))
graphics::par(mar = c(5.1, leftmargin, 4.1, 4.1))
```

PR #288, opened 2020-09-12
addresses #277 by truncating query names and expanding left margin

Aariq:
Old plot:
image

New plot: 
image


R/integration.R
```diff
@@ -120,14 +120,18 @@ find_db <- function(query, from,
         select(all_of(colorder)) %>%
         as.matrix()
       opar <- graphics::par(no.readonly = TRUE)
-      graphics::par(mar = c(5.1, 7.1, 4.1, 4.1)) # adapt margins
+
+      query_trunc <- str_trunc(query, 25)
```
stitam, line 124: I think `stringr` is not imported for the function, maybe `stringr::str_trunc(query, 25)`?

stitam (changes requested):
Thanks @Aariq. You're right, most compounds have long names, so truncating them is important, good idea! I will soon open a new issue to address "importing" in a wider context.

stitam (approved):

PR #296, opened 2020-09-29
Dev minor changes

andschar:
This is a minor PR, which changes `x == FALSE` to `!x` and `x == TRUE` to `x`  and replaces `URLencode()` in pubchem.R to not interfere with query names. See slack discussion.

### DIFF
--- a/R/pubchem.R
+++ b/R/pubchem.R
@@ -124,13 +124,13 @@ get_cid <-
            first = NULL,
            ...) {
 
-    if (ping_service("pc") == FALSE) stop(webchem_message("service_down"))
+    if (!ping_service("pc")) stop(webchem_message("service_down"))
 
   #deprecate `first`
-  if (!is.null(first) && first == TRUE) {
+  if (!is.null(first) && first) {
     message("`first = TRUE` is deprecated. Use `match = 'first'` instead")
     match <- "first"
-  } else if (!is.null(first) && first == FALSE) {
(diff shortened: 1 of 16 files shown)


R/pubchem.R
```diff
@@ -178,14 +178,21 @@ get_cid <-
...
       if (verbose) webchem_message("query", query, appendLF = FALSE)
-      if (is.character(query)) query <- URLencode(query, reserved = TRUE)
       if (from %in% structure_search) {
         qurl <- paste("https://pubchem.ncbi.nlm.nih.gov/rest/pug",
-                      domain, from, query, "json", sep = "/")
-      }
-      else {
+                      domain,
+                      from,
+                      URLencode(as.character(query), reserved = TRUE),
```
stitam, line 185: Just a comment, I originally used `if (is.character(query)) query <- URLencode(query, reserved = TRUE)` because I was concerned that in some cases the API would require a numeric input and so converting it to a character string might break the function. I now tested it with some of the examples and it seems to be working properly, so in this case this is more compact. Thanks for the update, let's merge it, we can always revert if anyone finds an issue with this later.

stitam (approved):
Thanks @andschar for this edit!

R/pubchem.R
andschar: Hm, I think as long as it's anyway in the `qurl` sring it's a character string anyway. Passing through the body is a different story.

Issue #287, opened 2020-09-11
Function to mangae wait times


andschar:
As discussed, a suggestion for a function to handle wait times for webchem requests. Please comment on it. Please also add comments on the policies of different web services and general recommendations, though I think this is very server specific. I will update the table.

```r
webchem_sleep <- function(time = NULL,
                          type = c('API', 'scrape')) {
  type <- match.arg(type)
  if (is.null(time)) {
    if (type == 'API') {
      time <- 0.2
    }
    if (type == 'scrape') {
      time <- 0.4
    }
  } else {
    if (!is.numeric(time)) {
      stop('Set time is not numeric.')
    }
  }

  Sys.sleep(time)
}
```

Recommended wait times for different web services:

| web service | time (s) | source | note |
|-----------------------|-------------------|----------|---------|
| ChEBI | x | https://www.ebi.ac.uk/chebi/webServices.do |
| Chemspider | x | https://developer.rsc.org/docs/compounds-v1-trial/1/overview | 1000 requests/month |
| CIR  | ? |  ? |
| Flavornet | x | | currently down |
| NIST | x | | |
| OPSIN | x | ? | | 
| PAN | x | | |
| Pubchem | 0.2 | https://pubchemdocs.ncbi.nlm.nih.gov/pug-rest |
| SRS | x | https://cdxnodengn.epa.gov/cdx-srs-rest/v2/api-docs | |
| Wikidata | x | https://www.wikidata.org/robots.txt | |

x...nothing stated
?...nothing found

Aariq:
SRS has nothing listed in API docs
OPSIN has nothing listed in robots.txt

Aariq:
PAN has nothing in docs or on robots.txt
Flavornet is down, but very unlikely that anything rate limits are documented as this is a small resource.
Nothing for NIST as far as I can find

andschar:
So I think we go for 0.2 seconds wait default, since we haven't found any other value. Ok?

PR #298, opened 2020-10-01
Final edits before v1.1

stitam:
* Added a startup message that points to the survey, let me know if you like it.
* Removed authors from function documentation, if any authors have an issue with this, we can revert later.
* Removed contributors from NEWS.md, if any contributors have an issue with this, we can revert later.

Any other quick fixes before we release v1.1?

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -25,3 +25,4 @@ README_cache
 ^\.github$
 
 codemeta.json
+^codemeta\.json$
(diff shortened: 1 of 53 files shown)


andschar:
Should we also include the wait time function from #287? I could file a PR for that and we have had a look at all the web services  concerning their policies, though mostly without finding anything.

stitam:
> Should we also include the wait time function from #287? I could file a PR for that and we have had a look at all the web services concerning their policies, though mostly without finding anything.

Yeah, if you're almost ready with it, then sure, let's put it in. I'll merge this PR last so it is really "final edits":)

andschar:
I have created PR #300

stitam:
> I have created PR #300

Just flagging within that we have agreed within the PR discussion that we'll not include this PR in v1.1.

PR #300, opened 2020-10-01
Webchem sleep function

andschar:
Here's the PR for a single function to sleep them all ;)
After #287. Let me know what you think.


andschar:
hm, don't really see what s happening here.

stitam:
> hm, don't really see what s happening here.

Tests work locally..

stitam:
So the tests seem to fail consistently for `cir`. Could it be because of the reduced wait times? I am now wondering if we should include this PR in v1.1. Maybe it would be worth to keep it in a dev version for a while to see if any other functions misbehave?

andschar:
> Maybe it would be worth to keep it in a dev version for a while to see if any other functions misbehave?

I agree let's wait a bit since this effects every function!

PR #302, opened 2020-10-06
Streamlining ChemSpider functions

stitam:
This PR significantly changes the internal functions for accessing ChemSpider.

Note, it does not change the user experience for any of the exported functions.

* `get_csid()` used various internal functions depending on the value of the `from` argument. This is because ChemSpider uses many APIs. These internal functions had very similar structures so I removed the redundancy and pooled them into `get_csid()`. `get_csid()` is now more complex, but it no longer communicates with other internal functions e.g. `get_name_csid()`, `get_formula_csid()`. These functions have become obsolete so I removed from `webchem`.
* `cs_convert()` also used various internal functions depending on the value of the `from` and `to` arguments. Similarly to `get_csid()` these internal functions were pooled into `get_convert()`. Internal functions like `cs_convert_multiple()` are now removed from `webchem`.

Other updates:

* `httr::RETRY()` and harmonised verbose messages.
* fail quickly when webservice is down.
* remove `@import`, `@importFrom`.

PR task list:
- [x] Update NEWS - probably not relevant, no change to user experience. retry and verbose messages are already in news.
- [x] Check package passed


stitam:
I'll fix the commit message where I accidentally referenced a GitHub user:)

stitam:
> I'll fix the commit message where I accidentally referenced a GitHub user:)

Done

R/chemspider.R
```diff
@@ -185,22 +188,106 @@ get_csid <- function(query,
   }
   from <- match.arg(from)
   match <- match.arg(match)
-
+  if (!ping_service("cs")) stop(webchem_message("service_down"))
```
Aariq, line 191: it just occurred to me that `ping_service()` could have a `verbose` argument that would just call `webchem_message("service_down")` if the service is down..

R/chemspider.R
```diff
@@ -656,33 +376,79 @@ cs_convert <- function(query, from, to, apikey = NULL) {
...
+      if (verbose) message("Conversion not supported. Returning NA.")
+      return(NA_integer_)
     }
+    else {
+      out <- get_csid(query, from = from, apikey = apikey)$csid
+      return(out)
+    }
+  }
+  if (from == "inchikey" & to == "smiles") {
+    if (verbose) message("Conversion not supported. Returning NA.")
```
Aariq, line 392: Just curious why you decided to make these messages and not errors. I would expect a combination of arguments that is not supported to result in an error or warning in most cases.

Aariq (commented):
I took a quick look through on GitHub (didn't test anything out in R) and it looks good to me.  I'm mostly just curious why you decided to change the errors to messages in `cs_convert()`.  I think it's probably fine, but to me it makes more sense for this to be a warning or error.  If a function can't do the thing it's being asked to do, it seems like at the very least they should be warned.  I tend to use these guidelines for error, warning, or message: https://adv-r.hadley.nz/conditions.html#signalling-conditions

R/chemspider.R
stitam: We had a discussion earlier about how functions should behave when the webservice is down. We concluded that in such a situation the function should fail fast. On second thought, you're right, this is a very similar situation and the function should return an error. Thanks for pointing this out.

R/chemspider.R
stitam: Thanks @Aariq for this suggestion, I like the idea, but I think it would be better to open a separate PR for this.

PR #306, opened 2020-10-13
vcr demo

maelle:
following discussion with @stitam

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto
+tests/fixtures/**/* -diff
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -50,6 +50,7 @@ Suggests:
     knitr,
     rmarkdown,
     plot.matrix,
-    usethis
+    usethis,
+    vcr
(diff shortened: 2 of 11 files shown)


Aariq:
Hey @stitam, I used `vcr` for tests in another project and have a decent handle on how it works.  Are we still interested in implementing it for testing `webchem`?  The downside is that changes to APIs won't cause our tests to fail (unless we keep some tests that query the API).

stitam:
Hi @Aariq, it takes forever to pass all the checks before merging a PR so I think vcr would be useful. To manage the downside, do you think e.g. manually rebuilding the vcr records every now and then would be feasible? I doubt this "every now and then" could be automated :/

maelle:
there is also {httptest} , alternatively. It's more actively maintained, and if you were to switch to httr2, switching your tests to {httptest2} would be easier than from {vcr} to {httptest2}.

Aariq:
Maybe I'll try out `httptest` for my other, simpler, project and report back.

maelle:
For info https://books.ropensci.org/http-testing/httptest.html :innocent:

PR #312, opened 2021-02-01
Use names instead of stats::setNames

stitam:
webchem used three functions from the `stats` package, `rexp()`, `rgamma()`, `setNames()`. The first two were removed with PR #300 so now I'm proposing to remove `setNames` as well to reduce the number of packages webchem depends on.

PR task list:
- [x] Check package passed


Aariq:
This is fine, but I wouldn't feel pressured to do anything to remove a dependency on `stats`. `stats` is part of base R and just about everyone who has installed R should have `stats` installed too.

PR #319, opened 2021-02-20
Update ChemSpider tests to use vcr 

stitam:
This PR adds the `vcr` package which allows recording http requests. I frequently run out of my monthly allowance at ChemSpider because of running all the ChemSpider tests locally before opening a PR and I figured vcr may help here. I think this change will also increase test coverage, as ChemSpider tests will no longer be skipped. Thanks for the lead @maelle!

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto
+tests/fixtures/**/* -diff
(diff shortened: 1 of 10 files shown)


maelle:
Awesome!

PR #336, opened 2021-11-15
fix ping_pubchem_pw() #335

### LINKED ISSUE #335: `ping_pubchem_pw()` returning FALSE when pubchem is not down
Looks like something changed with PUG VIEW.  The qurl used in `ping_pubchem_pw()` is no longer working ("https://pubchem.ncbi.nlm.nih.gov/rest/pug_view/data/compound/176/JSON?heading=pka").  If you just query the whole compound, it works though ("https://pubchem.ncbi.nlm.nih.gov/rest/pug_view/data/compound/176/JSON").  Might also change the compound to the one used in their documentation—1234?

>Hi,
>I get the following error when I try this command:
>
>```
>Service not available. Returning NA.
>Error in get_cid("C26H52NO6P", from = "formula")
>```

_Originally posted by @apuhegde in https://github.com/ropensci/webchem/issues/206#issuecomment-969314326_

Aariq:
Changes the URL used to check if PUG VIEW service is up (closes #335).  Also changes an expected response in test-pubchem.R (minor change in pubchem database entry).

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -7,7 +7,7 @@ Description: Chemical information from around the web. This package interacts
     Flavornet, NIST Chemistry WebBook, OPSIN, PAN Pesticide Database, PubChem,
     SRS, Wikidata.
 Type: Package
-Version: 1.1.1.9003
+Version: 1.1.1.9004
 Date: 2021-02-07
 License: MIT + file LICENSE
 URL: https://docs.ropensci.org/webchem/, https://github.com/ropensci/webchem
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -7,7 +7,7 @@ Description: Chemical information from around the web. This package interacts
     Flavornet, NIST Chemistry WebBook, OPSIN, PAN Pesticide Database, PubChem,
     SRS, Wikidata.
 Type: Package
-Version: 1.1.1.9002
+Version: 1.1.1.9003
 Date: 2021-02-07
 License: MIT + file LICENSE
 URL: https://docs.ropensci.org/webchem/, https://github.com/ropensci/webchem
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -7,7 +7,7 @@ Description: Chemical information from around the web. This package interacts
     Flavornet, NIST Chemistry WebBook, OPSIN, PAN Pesticide Database, PubChem,
     SRS, Wikidata.
 Type: Package
-Version: 1.1.1.9001
+Version: 1.1.1.9002
 Date: 2021-02-07
 License: MIT + file LICENSE
 URL: https://docs.ropensci.org/webchem/, https://github.com/ropensci/webchem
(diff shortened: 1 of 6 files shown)


Aariq:
So turns out Pubchem changed the name of the page with pKa on it, which is probably what caused all this.  One pubchem test is still failing, but it's not related to this change, I think.

R/ping.R
```diff
@@ -207,7 +207,7 @@ ping_pubchem <- function(...) {
 #'  }
 ping_pubchem_pw <- function(...) {
   qurl <- paste("https://pubchem.ncbi.nlm.nih.gov/rest/pug_view/data",
-               "compound/176/JSON?heading=density", sep = "/")
+               "compound/176/JSON", sep = "/")
```
stitam, line 210: I wonder, will the function now query the whole page, and will this make it slower to ping the service? What do you think? On the other hand, I like that this change would make `ping_pubchem_pw()` robust to further changes in the structure of a PubChem page.

tests/testthat/test-pubchem.R
```diff
@@ -114,13 +114,13 @@ test_that("pc_page()", {
...
 
-  a <- pc_page(c(311, 176, 1118, "balloon", NA), "pKa")
+  a <- pc_page(c(311, 176, 1118, "balloon", NA), "Dissociation Constants")
 
   expect_type(a, "list")
   expect_length(a, 5)
   expect_s3_class(a[[1]], c("Node", "R6"))
   expect_s3_class(a[[2]], c("Node", "R6"))
-  expect_equal(a[[3]], NA)
+  expect_equal(a[[3]], NA) #why was this expected to be NA?
```
stitam, line 123: Hard to believe sulfuric acid did not have pka values on PubChem but I do remember that all tests used to pass.. Anyway it is no longer NA so I think you can remove this test.

DESCRIPTION
```diff
@@ -7,7 +7,7 @@ Description: Chemical information from around the web. This package interacts
     Flavornet, NIST Chemistry WebBook, OPSIN, PAN Pesticide Database, PubChem,
     SRS, Wikidata.
 Type: Package
-Version: 1.1.1.9004
+Version: 1.1.1.9003
```
stitam, line 10: Can you please bump it above 9004? I am guessing we started working on this issue at the same time :)

stitam (changes requested):

R/ping.R
Aariq: It might be slightly slower, but I think its worth it to be less fragile.

DESCRIPTION
Aariq: Yeah, this was because the version in the NEWS file was different from that in the DESCRIPTION.  I used `usethis::use_version(which = "dev")` , in this PR, which handles both.  I'll bump it to 9004.

R/ping.R
Aariq: Functions should only `ping()` once each time they're run, right?  The speed change should be minimal.

sneumann:
I can confirm that this branch installs on linux here and fixes our previously failing `pc_prop(96)` since the ping now succeeds. Thanks, yours, Steffen

PR #339, opened 2021-11-18
The alanwood ISO common name site has moved

jranke:
Brief description of the PR

As mentioned above, https://alanwood.net/pesticides has moved. Unfortunately, the new server results in an error with default curl options, so I had to make sure that `accept_encoding` does not include `gzip`.

PR task list:
- [x] Update NEWS
- [x] Tests were updated
- [ ] Check package passed


Aariq:
Nice work.  My first thought is that we should not remove `aw_*()` functions immediately.  They could be soft deprecated by doing something like:

```
aw_query <- function( query,
  from = c("name", "cas"),
  verbose = getOption("verbose"),
  type,
  ...) {
message("`aw_query()` will be deprecated in the next version of webchem.  Please update code to use `bcpc_query()` instead.")
bcpc_query(query = query, from = from, verbose = verbose, type = type, ...)
}
```
There might be a better way to implement it, but basically `aw_query()` should still work by just calling the new function `bcpc_query()` and should warn the user to start using `bcpc_query()` instead.

Then, in the next version, `aw_query()` will error, pointing user to `bcpc_query()` instead.  And in the next, *next* version of webchem, we can remove the `aw_*` functions entirely.

R/webchem-deprecated.R
```diff
@@ -10,9 +10,18 @@
 #' @aliases cid_compinfo
 #' @details Deprecated functions are:
 #' \tabular{rl}{
-#'   \code{pc_compinfo} \tab is now a synonym for \code{\link{cid_compinfo}}\cr
+#'   \code{pc_prop} \tab was formerly \code{\link{cid_compinfo}}\cr
+#'   \code{bcpc_query} \tab was formerly \code{\link{aw_query}}\cr
 #' }
 cid_compinfo <- function(...) {
-  .Deprecated("pc_compinfo", package = "webchem")
+  .Deprecated("pc_prop", package = "webchem")
```
Aariq, line 17: Can you explain this change?

Aariq:
> Nice work. My first thought is that we should not remove `aw_*()` functions immediately. They could be soft deprecated by doing something like:

Ah wait, that is what `.Deprecated()` does, isn't it.  Sorry, didn't see that before.

R/webchem-deprecated.R
jranke: We do not have `pc_compinfo` any more (see NAMESPACE and R/pubchem.R), AFAICT this is now provided by `pc_prop`. So the suggestion has to be adapted.

jranke:
> Nice work. My first thought is that we should not remove `aw_query()` immediately. They could be **soft deprecated** by doing something like:

...

Yes, at the moment if you accept this PR `aw_query` will give a warning but still works and gives results.

Aariq:
> Yes, at the moment if you accept this PR `aw_query` will give a warning but still works and gives results.

Thanks! I forgot how `.Deprecated()` works 😳

R/integration.R
```diff
@@ -73,15 +73,15 @@ with_cts <- function(query, from, .f, .verbose = getOption("verbose"), ...) {
 #' }
 find_db <- function(query, from,
                            sources = c("etox", "pc", "chebi", "cs",
-                                       "aw", "fn", "pan", "srs"),
+                                       "bcpc", "fn", "pan", "srs"),
```
Aariq, line 76: Do we want to soft deprecate here as well (e.g. include both "aw" and "bcpc" as synonymous options)?  I think this change is probably fine since it is a relatively new function anyways and likely not part of many workflows, but just want to hear what @stitam or @andschar think.

Aariq:
Could you also update README.Rmd to change the link to Alanwood

jranke:
> Could you also update README.Rmd to change the link to Alanwood

I already did:

https://github.com/ropensci/webchem/pull/339/files?file-filters[]=.Rmd

Aariq:
Going to go ahead and merge this. I think the change to `find_db()` is fine

PR #343, opened 2021-12-12
Update pkgdown

Aariq:
updates _pkgdown.yml to reflect move of Alanwood database and addition of `write_mol()`.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,5 +1,10 @@
 url: https://docs.ropensci.org/webchem
 
+template:
+  package: rotemplate
+
+destination: docs
+
 reference:
 - title: Retrieve and translate chemical identifiers
 - contents:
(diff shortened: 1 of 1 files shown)


Aariq:
hmm... not sure why codecov isn't running

_pkgdown.yml
```diff
@@ -15,9 +20,8 @@ reference:
   - etox_basic
 - title: Retrieve chemical properties
 - contents:
-  - aw_query
-  - ci_query
-  - cir_query
+  - ends_with("_query")
+  - -opsin_query
```
stitam, line 24: I'm wondering, is there a typo here?

stitam (approved):
Hi @Aariq thanks for the fix! I added a comment.

_pkgdown.yml
Aariq: `opsin_query()` is a identifier translator, so I left it out of this section on purpose.  It shows up in the chemical identifier translator section.

Issue #283, opened 2020-09-04
get_cid fails with search via SMILES in Pubchem


stephanflemming:
Some of my queries for SMILES in Pubchem result in NA, while the Pubchem search is able to find that compound.

Example: O=S(=O)(NCCNC/C=C/c1ccc(Br)cc1)c2cccc3cnccc23

https://pubchem.ncbi.nlm.nih.gov/compound/449241

and 

```
> get_cid("O=S(=O)(NCCNC/C=C/c1ccc(Br)cc1)c2cccc3cnccc23", from = "smiles")
Querying O=S(=O)(NCCNC/C=C/c1ccc(Br)cc1)c2cccc3cnccc23. Bad Request (HTTP 400).
# A tibble: 1 x 2
  query                                         cid  
  <chr>                                         <lgl>
1 O=S(=O)(NCCNC/C=C/c1ccc(Br)cc1)c2cccc3cnccc23 NA  
```

I am using the development version 1.0.0.9000.

Thanks in advance!

stitam:
Hi @stephanflemming thanks for raising this issue.

When I search for "O=S(=O)(NCCNC/C=C/c1ccc(Br)cc1)c2cccc3cnccc23" on PubChem (so not through webchem), it finds CID 449241, but according to the PubChem page of CID 449241, the smiles of the compound is "C1=CC2=C(C=CN=C2)C(=C1)S(=O)(=O)NCCNCC=CC3=CC=C(C=C3)Br". I am unsure why the two smiles IDs don't match. When I search for "C1=CC2=C(C=CN=C2)C(=C1)S(=O)(=O)NCCNCC=CC3=CC=C(C=C3)Br" I get CID 3541, and I don't get CID 449241. However the two compounds, CID 449241 and CID 3541 appear to be the same. This is strange, I am unsure what's happening.

Regardless, our understanding has been that `get_cid()` returns the exact same output as the search bar on PubChem. Based on your finding, either the search bar is more sophisticated that that or `get_cid()` does something wrong. We'll work on it.

If you could provide more examples, that would be great.

Aariq:
If you look closely at the web API search results, it says "Find structures very closely related to the input, comparing chemical connectivity, and optionally tautomers, stereoisomers, and isotopes."  So I think the web search is doing something slightly different and returning compounds that are very close matches to the query.

schymane:
[...]

PR #346, opened 2022-01-27
Fix url construction for smiles queries

stitam:
This PR is linked to Issue #283.

First I tried to update the original url constructor based on the url suggested within Issue #283, like so:

```
qurl <- paste0("https://pubchem.ncbi.nlm.nih.gov/rest/pug/",
                       domain, "/", 
                       from, "/", 
                       "cids/JSON?",
                       from, "=",
                       URLencode(as.character(query), reserved = TRUE))
```

Notice the small change: after `cids/JSON?` I substituted `smiles=` ,with `from, "="` to generalise. However, many of the tests broke. Therefore, I just added the suggested url constructor as a special case -> `from == "smiles"`. What do you think?

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -28,3 +28,4 @@ codemeta.json
 ^codemeta\.json$
 
 cran-comments.md
+^CRAN-RELEASE$
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+# webchem 1.1.2.9001
+
+## BUG FIXES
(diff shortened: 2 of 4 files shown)


Aariq:
Are we sure SMILES the only identifier that can have with `/` in it?  If so, then just making `from="smiles"` a special case seems fine.

Aariq (approved):

Issue #289, opened 2020-09-15
return tibble from cir_query()


andschar:
`cir_query()` currently returns a list of vectors. Should we return a `tibble` instead? I know it's not a `get_*()` function, but it behaves very similar to them, since only identifiers are returned. Then again, it would be a breaking change.

Aariq:
related to #218

Aariq:
Shouldn't break anything, I think, since a tibble *is* a list of vectors.

PR #350, opened 2022-02-15
Change `cir_query()` output to tibble #289

Aariq:
This changes the output of `cir_query()` to a tibble, a potentially breaking change for some.  Suggested in #289.  I also made a small fix to the `matcher()` utility so the query is printed when match = "ask".

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/R/cir.R
+++ b/R/cir.R
@@ -131,11 +131,11 @@ cir_query <- function(identifier,
   }
 
   match <- match.arg(match)
-
+  na_tbl <- tibble(query = NA, !!representation := NA)
   foo <- function(identifier, representation, resolver, match, verbose) {
     if (is.na(identifier)) {
       if (verbose) webchem_message("na")
-      return(NA)
+      return(na_tbl)
     }
     if (verbose) webchem_message("query", identifier, appendLF = FALSE)
     identifier <- URLencode(identifier, reserved = TRUE)
@@ -152,7 +152,7 @@ cir_query <- function(identifier,
                          quiet = TRUE), silent = TRUE)
     if (inherits(h, "try-error")) {
       if (verbose) webchem_message("service_down")
-      return(NA)
+      return(na_tbl)
     }
     if (verbose) message(httr::message_for_status(h))
     if (h$status_code == 200){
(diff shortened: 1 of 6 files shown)


stitam (changes requested):
Thanks @Aariq for this PR, it looks good! One thing I noticed, when the query cannot be found in the database, the function returns `NA` in the query column as well (e.g. `cir_query("chemical", "cas")`). Do you see this on your end?

Aariq:
Yeah.  This was on purpose.  I think it's consistent with how the `get_*()` functions work?  It's useful for doing things like:

``` r
library(tidyverse)
library(webchem)
df <- tibble(cas = c("3380-34-5", NA, "56-40-6"))

df %>% 
  mutate(smiles = cir_query(cas, "smiles", match = "first")$smiles)

#> # A tibble: 3 × 2
#> cas       smiles                      
#> <chr>     <chr>                       
#>   1 3380-34-5 Oc1cc(Cl)ccc1Oc2ccc(Cl)cc2Cl
#> 2 NA        NA                          
#> 3 56-40-6   NCC(O)=O 
```

stitam:
Thanks @Aariq, I agree when the query is `NA` the output should include a row where the query column is also `NA`. But when the query is not `NA` but a string which cannot be found in the database, I think the functions should output the query string in the query column, like here:

``` r
webchem::get_cid(c("gentamycin", "pumpkin"))
#> # A tibble: 5 × 2
#>   query      cid      
#>   <chr>      <chr>    
#> 1 gentamycin 3467     
#> 2 gentamycin 72395    
#> 3 gentamycin 441305   
#> 4 gentamycin 132477936
#> 5 pumpkin    <NA>
```

<sup>Created on 2022-02-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

When I tested this with the new `cir_query()` I get `NA` in the query column as well: 

`cir_query(c("gentamycin", "pumpkin"), "smiles")`

```r
# A tibble: 2 × 2
  query      smiles                                                   
  <chr>      <chr>                                                    
1 gentamycin CNC(C)C1CCC(N)C(O1)OC2C(N)CC(N)C(OC3OCC(C)(O)C(NC)C3O)C2O
2 NA         NA  
```

Aariq:
Ooooh, gotcha.  I'll fix this.

R/cir.R
```diff
@@ -19,7 +20,7 @@
 #' @param choices deprecated.  Use the \code{match} argument instead.
 #' @param verbose logical; should a verbose output be printed on the console?
 #' @param ... currently not used.
-#' @return A list of character vectors.
+#' @return A tibble with a `query` column an a column for the requested representation.
```
stitam, line 23:
```suggestion
#' @return A tibble with a `query` column and a column for the requested representation.
```

stitam (approved):
Thanks @Aariq.

PR #353, opened 2022-03-29
Functions for accessing a new database - ChEMBL

stitam:
This PR adds new functionality for accessing the ChEMBL database (https://www.ebi.ac.uk/chembl/). The database developers
created an API for retrieving the data (https://chembl.gitbook.io/chembl-interface-documentation/web-services), which I think evaluates to explicit consent to programmatic access, so we should be fine.

I would like to point out that I also added some caching functionality to `chembl_query()`. This was before Discussion #352. I am actively using this functionality, but I am OK to remove it if you think we should first discuss caching in more detail. Alternatively, I can merge this, implement other solutions with other webchem functions and then we can discuss/settle with one and update it everywhere. What do you think?

PR task list:
- [x] Update NEWS
- [x] Check package passed


stitam:
More info on licencing: https://chembl.gitbook.io/chembl-interface-documentation/about#chembl-team-members

stitam:
I have rerun the tests several times, `with_cts()` works on my end, I don't know why the ubuntu release version is not working..

Aariq:
I'm happy to review this, but it won't happen until next week at the earliest, FYI.  I'm excited to see how you made caching work!

stitam:
Thanks perfect, thanks @Aariq!

Aariq:
Hey, I just realized I never took a look at this.  I'll try and take a look sometime in the next few weeks

stitam:
No worries, thanks :)

R/chembl.R
```diff
@@ -0,0 +1,226 @@
+#' Query ChEMBL using ChEMBL IDs
+#'
+#' Retrieve ChEMBL data using a vector of ChEMBL IDs.
+#' @param query character; a vector of ChEMBL IDs.
+#' @param resource character; the ChEMBL resource to query. Use
+#' \code{chembl_resources()} to see all available resources.
```
Aariq, line 6: link to chembl_resources() help page
```suggestion
#' [chembl_resources()] to see all available resources.
```

R/chembl.R
```diff
@@ -0,0 +1,226 @@
...
+  resources <- c(
+    "activity", "activity_supplementary_data_by_activity", "assay",
+    "assay_class", "atc_class", "binding_site", "biotherapeutic", "cell_line",
+    "chembl_id_lookup", "compound_record", "compound_structural_alert",
+    "document", "document_similarity", "drug", "drug_indication",
+    "drug_warning", "go_slim", "image", "mechanism",  "metabolism",
+    "molecule", "molecule_form", "organism", "protein_class", "similarity",
+    "source", "status", "substructure", "target", "target_component",
+    "target_relation", "tissue", "xref_source"
+  )
```
Aariq, line 197: Any way to get this from the API?  Just trying to think ahead to when they add another resource

R/chembl.R
```diff
@@ -0,0 +1,226 @@
...
+#' contains a list of respective query results. Results are simplified, if
+#' possible.
+#' @details Each entry in ChEMBL has a unique ID. Data in ChEMBL is organized in
+#' databases called resources. An entry may or may not have a record in a
+#' particular resource. An entry may have a record in more than one resource,
+#' e.g. a compound may be present in both the "molecule" and the "drug"
+#' resource. This function queries a vector of ChEMBL IDs from a specific ChEMBL
+#' resource. If you are unsure which ChEMBL resource contains your ChEMBL ID,
+#' use this function with the \code{"chembl_id_lookup"} resource to find the
+#' appropriate resource for a ChEMBL ID.
```
Aariq, line 21: Can you add an example of how to do this?

R/chembl.R
```diff
@@ -0,0 +1,226 @@
...
+#' \donttest{
+#' # Might fail if API is not available
+#'
+#' atc <- atc_classes()
+#' }
+#' @importFrom httr RETRY user_agent message_for_status content
+#' @importFrom tibble as_tibble
+#' @importFrom dplyr bind_rows
+#' @export
+atc_classes <- function(verbose = getOption("verbose"),
```
Aariq, line 140: Should this be called `chembl_atc_classes()`?  Or is this something separate from ChEMBL?

Aariq (changes requested):

R/chembl.R
stitam: So far I haven't found a way to get this from the API, but I can add at least one test for each resource. Though we won't notice when a new resource is added, but we will notice if any of the existing resources are changed/removed.

R/chembl.R
stitam: I wasn't sure because the ATC system itself is a general thing but you're right, the data set is queried from ChEMBL so it should be called `chembl_atc_classes()`, thanks.

R/chembl.R
stitam: There is an example in the examples section but maybe it's not visible enough? I've updated the function documentation a bit, let me know if it's still not clear enough.

Aariq: Yeah, looks good to me

Aariq (approved):

PR #356, opened 2022-04-21
Update cts_convert to accept numeric input

stitam:
`cts_convert()` produced an error when the query was not a string. However, in certain cases, like with PubChem CIDs, numeric queries are perfectly valid. This PR suggests and edit to the `cts_convert()` function: queries are automatically converted to characters.

PR task list:
- [ ] Update NEWS
- [ ] Check package passed

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
 
 ## BUG FIXES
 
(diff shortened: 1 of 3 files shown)


Aariq (approved):
Makes sense to me!

Issue #374, opened 2022-10-25
RI Extraction issue


Ellakai:
Dear developper,

Thanks agin for your script :)
very nice when we want extract big set of information.

Now, i would like to extract RI from nist.
In the past, i tried and it was running ; but now i would like to use in real case and i obtain always the same error message :(

For exemple, if i use this code

`CAS <- c("431-03-8")
RI.table <- nist_ri(CAS,
                    from = "cas",
                    type = "linear",
                    polarity = "non-polar",
                    temp_prog = "ramp")`

We can see on the website lot or RI available : https://webbook.nist.gov/cgi/cbook.cgi?ID=C431038&Units=SI&Mask=2000#Gas-Chrom

but i obtain always this message

`Error in `dplyr::select()`:
! Can't subset columns that don't exist.
✖ Column `Phase thickness (m)` doesn't exist.
Run `rlang::last_error()` to see where the error occurred.`

and, if i try
`CAS <- c("431-03-8")
RI.table <- nist_ri(CAS,
                    from = "cas",)`

I obtain en empty table.

Thanks for your help :)
regards

Aariq:
Thanks for the bug report! It looks like NIST maybe changed the phase thickness column title to "Phase thickness µm".  I'm guessing it was *always* in µm, but previous the "µ" wasn't there for some reason.  I'll double check that this change was consistent for all compounds and then make a change in the webchem code.

Ellakai:
thanks for your job :)
perfect if you found the issue ^^

thanks

Aariq:
This should be fixed now.  You can install the development version of webchem that includes the fix with the `remotes` package

``` r
remotes::install_github("ropensci/webchem")
```

PR #375, opened 2022-10-25
Fix for #374, adds "µ" to column name.

Aariq:
## Pull Request
Maybe the smallest PR in webchem history! Changes "Phase thickness (m)" to "Phase thickness (µm)" following an apparent change to the NIST website. (see #374 ).  Considered using something like `starts_with("Phase thickness")`, but I actually would prefer knowing about changes to things like this on NIST's end.

I also fixed a failing pubchem test.

I also updated the GH actions to use tagged versions of actions/cache, r-lib/actions/setup, etc.

PR task list:
- [x] Update NEWS
- [ ] Check package passed

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -40,7 +40,7 @@ jobs:
         with:
           r-version: ${{ matrix.config.r }}
 
-      - uses: r-lib/actions/setup-pandoc@master
+      - uses: r-lib/actions/setup-pandoc@v2
 
       - name: Query dependencies
         run: |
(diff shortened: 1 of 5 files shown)


Aariq:
from check:
```
Warning: Found the following file with non-ASCII characters:
  nist.R
Portable packages must use only ASCII characters in their R code,
except perhaps in comments.
Use \uxxxx escapes for other characters.
```

Aariq:
The failing test also fails when I run `devtools::test()` locally, but not when I click the "Run Tests" button in RStudio.  Any ideas?

Aariq:
Looks like CTS is not working consistently (not surprising).  Re-ran failing tests and now they all work.

PR #379, opened 2022-11-05
Return CAS number from nist_ri

ethanbass:
This PR contains implements suggested changes to the `nist_ri` function so that it returns an additional column containing the CAS number along with the other retention index information. Currently, if you use a query other than the CAS number, the function queries NIST for the CAS number and prints it to the console, but it is not recorded. It is especially useful to have the CAS number if you want to get multiple kinds of retention indices (e.g. kovats and alkane or polar and non-polar) for the same query. With the altered function, you can use the CAS number returned by the first query to request additional information, instead of querying NIST for the CAS number on each request.

PR task list:
- [x ] Update NEWS
- [ x] Add tests (if appropriate) (NA)
- [ x] Check package passed


R/nist.R
```diff
@@ -153,7 +154,7 @@ get_ri_xml <-
 tidy_ritable <- function(ri_xml) {
   #Skip all these steps if the table didn't exist at the URL and was set to NA
   if (any(is.na(ri_xml))) {
-    return(tibble(RI = NA))
+    return(tibble(RI = NA, cas=attr(ri_xml, "cas")))
```
stitam, line 157: I am not sure about this one because:

``` r
nist_ri("balloon", verbose = TRUE)
#> Querying balloon. OK (HTTP 200).
#> Not available. Returning NA.
#> # A tibble: 1 × 2
#>   query   RI      cas   
#>   <chr>   <lgl>   <chr>
#> 1 balloon NA      balloon
```

I think in this case "cas" should also be NA, what do you think?

My other thought is just a matter of taste, I am wondering if it would look better if the table returned by `nist_ri()` contained first the query, then cas, then RI, then everything else? It feels tidier to start with identifiers

R/nist.R
```diff
@@ -252,6 +253,7 @@ tidy_ritable <- function(ri_xml) {
           TRUE                  ~ as.character(NA)
         )
       ) %>%
+      dplyr::mutate(cas = attr(ri_xml, "cas")) %>%
       # reorder columns
       dplyr::select("RI", "type", "phase", everything())
```
stitam, line 258: My comment above may have an impact here as well.

stitam (changes requested):
Many thanks @ethanbass for opening this PR! I agree it would be helpful to include cas in the function output. A new version of webchem has been released recently so I moved your NEWS entry up a little. I added a few comments, can you please resolve these?

R/nist.R
Aariq: Yeah, should be query, cas, RI I think.

Aariq:
It would be great to have tests for the following:

- a name query that returns a CAS number
- a name query for something without RI data, but that still has a CAS number (e.g. methane)
- a name query for something with RI data but without a CAS number, e.g. https://webbook.nist.gov/cgi/cbook.cgi?ID=R618870&Units=SI
- Is there any situation where a search for a CAS number would return a different CAS number? If that's a thing (honestly can't remember), we should have a test that that works correctly too maybe?

If you need help adding tests, let me know and I can help or point you toward examples

ethanbass:
Sounds good. I can try to write some tests. For scenario 2, the name that doesn't have RI data, (e.g. methane), this type of query will not return a CAS number at all the way the function is currently written. What do you all think the expected behavior should be here? I think the function would need to be significantly modified to make it return a cas number for this type of scenario.

For scenario 4, I can't think of an example where a cas number will return a different cas number, since as far as I can tell there aren't any records with more than one cas number. I'd be happy to try to implement this test if anyone has an example though.

R/nist.R
stitam: There is a helper function `is.cas()` in webchem that validates cas strings, maybe that could help?

Aariq:
> Sounds good. I can definitely try to write some tests. For scenario 2, the name that doesn't have RI data, (e.g. methane), this type of query will not return a CAS number at all the way the function is currently written. What do you all think the expected behavior should be here? I think the function would need to be significantly modified to make it return a cas number for this type of scenario.

I think returning `NA` is probably OK as long as it's clear in the function documentation that the CAS number will be returned *if and only if* RI data is found.


> For scenario 4, I can't think of an example where a cas number will return a different cas number, since as far as I can tell there aren't any records with more than one cas number. I'd be happy to try to implement this test if anyone has an example though.

I think you're probably right.  For some reason I was thinking NIST webbook redirected isomers or something, but played around with it a bit this morning and that doesn't seem to be the case.  Let's not worry about this

R/nist.R
ethanbass: I was thinking more about this. Do ya'll think it would make sense to have an additional logical argument (e.g. `validate_cas`)? If set to TRUE, the function could check that the CAS number is valid. If false, it will just assume that the CAS number is valid which is the current behavior. Could be still set to FALSE by default?

Aariq:
If I remember correctly, right now it's actually scraping the "large format table" (e.g. https://webbook.nist.gov/cgi/cbook.cgi?ID=C67561&Units=SI&Mask=2000&Type=LINEAR-RI-NON-POLAR-CUSTOM#LINEAR-RI-NON-POLAR-CUSTOM) and these have different row names (which get converted to column names by webchem).  If you can think of a good way to harmonize these tables then it would be nice to be able to combine them all into one output.  I.e. allow something like `nist_ri('67-56-1', type =  c("kovats", "linear"), polarity = c("polar", "non-polar"), temp_prog = "ramp")` to download RIs for kovats + polar, kovats + non-polar, linear + polar, and linear + non polar with a temperature ramp.  I think I previously decided against doing this because the tables have different info (e.g. T_start and T_end only appear in temperature ramp tables)

stitam:
I pushed  a commit that changes the following: 1. when we call `nist_ri()` to look up cas numbers, the function will stop with an error message if any query cas is invalid, before connecting to the webservice. I think stopping at input validation is not against our rules for handling input validation: once we start contacting the webservice, we don't want any queries to stop the process but before doing that I suppose it's finem do you agree @Aariq? 2. in the final output table the cas is now formatted appropriately (I looked at code comments within `is.cas()` to determine the rules).

ethanbass:
> I pushed a commit that changes the following: 1. when we call `nist_ri()` to look up cas numbers, the function will stop with an error message if any query cas is invalid, before connecting to the webservice. I think stopping at input validation is not against our rules for handling input validation: once we start contacting the webservice, we don't want any queries to stop the process but before doing that I suppose it's finem do you agree @Aariq? 2. in the final output table the cas is now formatted appropriately (I looked at code comments within `is.cas()` to determine the rules).

This is a cool idea! one potential issue I see, is that some of the record identifiers used by the NIST webbook don't seem to beactual CAS numbers, e.g. "R628941" -- one of the strings you use as an example for the new `format_cas` function . The most straightforward way to get records for E-longifolene is to provide this string as the cas number, but as far as I can tell isn't an actual cas number and will thus be rejected by the `is.cas` function. For example, if you run:
```
query <- webchem:::format_cas("R628941")
is.cas(query)
```
It returns FALSE, which will now cause the nist_ri function to error out.

Aariq:
>I pushed a commit that changes the following: 1. when we call nist_ri() to look up cas numbers, the function will stop with an error message if any query cas is invalid, before connecting to the webservice. I think stopping at input validation is not against our rules for handling input validation: once we start contacting the webservice, we don't want any queries to stop the process but before doing that I suppose it's finem do you agree @Aariq?

I don't think this is consistent with other webchem functions. I think we more commonly have the function not query the API / do the scraping, but continue to run and return `NA` for that particular query with a warning.  In the future, we could consider making this behavior optional with some kind of `stop_on_fail` option or argument.

>2. in the final output table the cas is now formatted appropriately (I looked at code comments within is.cas() to determine the rules).

Nice! There is an exported `as.cas()` function already though.  Please use that unless there is some reason it won't work in this case.

>This is a cool idea! one potential issue I see, is that some of the record identifiers used by the NIST webbook don't seem to be actual CAS numbers -- e.g. "R628941"

I feel like I've had this discussion before about the non-CAS identifiers on NIST, but I don't remember what we did and I can't find any issues about it!  I think they should get returned as-is probably, but it might be worthwhile learning more about what they are.

R/nist.R
```diff
@@ -361,3 +371,24 @@ nist_ri <- function(query,
...
+format_cas <- function(x) {
+  # format CAS
+  x <- strsplit(gsub("^[a-zA-Z]+", "", x), "")[[1]]
+  len <- length(x)
+  first <- paste(x[1:(len-3)], collapse = "")
+  second <- paste(x[(len-2):(len-1)], collapse = "")
+  third <- x[len]
+  cas <- paste(first, second, third, sep = "-")
+  return(cas)
+}
```
Aariq, line 394: can you use the already existing `as.cas()` instead?

stitam: Oh I forgot... sure!

ethanbass:
> If I remember correctly, right now it's actually scraping the "large format table" (e.g. https://webbook.nist.gov/cgi/cbook.cgi?ID=C67561&Units=SI&Mask=2000&Type=LINEAR-RI-NON-POLAR-CUSTOM#LINEAR-RI-NON-POLAR-CUSTOM) and these have different row names (which get converted to column names by webchem). If you can think of a good way to harmonize these tables then it would be nice to be able to combine them all into one output. I.e. allow something like `nist_ri('67-56-1', type = c("kovats", "linear"), polarity = c("polar", "non-polar"), temp_prog = "ramp")` to download RIs for kovats + polar, kovats + non-polar, linear + polar, and linear + non polar with a temperature ramp. I think I previously decided against doing this because the tables have different info (e.g. T_start and T_end only appear in temperature ramp tables)

Ya that's the type of syntax I had in mind as well. I think I have solution that will work for this. I'll try to finish it and push something soon.

ethanbass:
It was more complicated than I thought, but I think I figured out a pretty good way to implement the syntax with multiple arguments. One potential issue is that it now defaults to collecting all of the available retention index information, which puts more load on the server because it has to send a different query for each of the long format tables. I'm not sure if it would be better to set the default more restrictively? I also added some more tests of the types that you suggested @Aariq

ethanbass:
I saw that the tests failed, but as far as I can tell the failing tests aren't ones that I changed. Is this related to anything that i did? I think not. If so, I can try to look into it though

Aariq:
Failing tests are very possibly not your fault.  I'll take a closer look at this PR today or tomorrow and let you know if there are any other changes needed.  Thanks for putting a lot of thought into how this should work!

stitam:
Thanks @ethanbass! The tests pass locally on my computer as well so this is probably just a temporary glitch. I'll restart the tests in a bit.

stitam (changes requested):
Thanks @ethanbass I see this PR grew quite a bit :D Thanks for improving the `format_cas()` function, I am now not sure it is needed at all :/ I have added a few comments, let me know what you think. Also feel free to add yourself as contributor.

R/nist.R
stitam: @ethanbass updated the function and I added a comment there so I am going to mark this one as resolved.

tests/testthat/test-nist.R
```diff
@@ -129,3 +147,35 @@ test_that("nist_ri works with NAs", {
...
+  expect_equal(test1[[1,"cas"]], "22374-89-6")
+  test2 <- nist_ri("α-methyl-Benzenepropanamine", from = "name", type="lee")
+  expect_equal(test2$query, "α-methyl-Benzenepropanamine")
+  expect_equal(test2$cas, "22374-89-6")
+})
+
+test_that("a record with no CAS number returns expected output", {
+  test <- nist_ri("(Z,Z)-alpha-Farnesene", from="name", type = "linear", polarity="polar")
+  expect_equal(test$query, "(Z,Z)-α-Farnesene")
+  expect_equal(test$cas, "U293031")
```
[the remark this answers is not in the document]
ethanbass: Yes, I wasn't too sure how to best deal with this. It's some kind of internal ID -- I don't really understand where these IDs come from. Usually they are the CAS number, but sometimes (as in this example) they're not. I'd be fine with just calling the column "nist_id" instead of "cas".

tests/testthat/test-nist.R
```diff
@@ -129,3 +147,35 @@ test_that("nist_ri works with NAs", {
...
+})
+
+test_that("nist_ri() can deal appropriately with a mixture of queries", {
+  test <- nist_ri(query=c(NA, "baloon", "methane", "hexanol"), from = "name",
+                  type=c("kovats","linear"), polarity="polar",
+                  temp_prog = "ramp", verbose=TRUE)
```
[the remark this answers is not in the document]
ethanbass: No I just left this in by accident. I'll fix it now

R/nist.R
```diff
@@ -252,112 +364,74 @@ tidy_ritable <- function(ri_xml) {
...
-        verbose = verbose
-      )
-    )
+format_cas <- function(ID){
+  format_cas <- function(ID){
+    cas <- as.cas(gsub("C","",ID))
+    ifelse(is.cas(cas) & !is.na(cas), cas, ID)
+  }
+ sapply(ID, format_cas)
+}
```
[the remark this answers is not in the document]
ethanbass: Personally I think it is cleaner and better to still have this as a separate function, but I agree that it should have the `noRd` tag. I will add it. It could also work to do the transformations on the spot, but it seems clunkier to me.

ethanbass:
> Thanks @ethanbass I see this PR grew quite a bit :D Thanks for improving the `format_cas()` function, I am now not sure it is needed at all :/ I have added a few comments, let me know what you think. Also feel free to add yourself as contributor.

Yes, I didn't quite realize what I was getting into... thanks for your suggestions! I went ahead and added myself to the description as suggested and resolved two of your comments. Still not quite sure what the best way is to deal with the weirdo CAS numbers. I'm think just changing the column name to `nist_id` might be the easiest/best approach, but i'm not sure

tests/testthat/test-nist.R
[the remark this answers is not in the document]
ethanbass: One more thought on this-- it also would not be hard to split the difference and return two columns here -- one with the "nist_id" and one only with true CAS numbers, however, they would be mostly redundant with each other.

R/nist.R
```diff
@@ -1,3 +1,128 @@
...
+#' @export
+#' @note Copyright for NIST Standard Reference Data is governed by the Standard
+#' Reference Data Act, \url{https://www.nist.gov/srd/public-law}.
+#' @seealso \code{\link{is.cas}} \code{\link{as.cas}}
+#'
+#' @examples
+#' \dontrun{
+#' myRIs <- nist_ri(c("78-70-6", "13474-59-4"), from = "cas", "linear",
+#' "non-polar", "ramp")
+#' }
```
Aariq, line 77: Let's show off the new feature in the example

```suggestion
#' myRIs <-
#'   nist_ri(
#'     c("78-70-6", "13474-59-4"),
#'     from = "cas",
#'     type = c("linear", "kovats"),
#'     polarity = "non-polar",
#'     temp_prog = "ramp"
#'   )
myRIs
```

tests/testthat/test-nist.R
```diff
@@ -32,7 +36,7 @@ test_that("nist_ri() returns results", {
   skip_on_cran()
   skip_if_not(up, "NIST Web Book is down")
 
-  out <- nist_ri("78-70-6", type = "linear", temp_prog = "custom")
+  out <- nist_ri("78-70-6", type = "linear", polarity="non-polar", temp_prog = "custom")
   expect_s3_class(out, "data.frame")
   expect_true(!anyNA(out$RI))
```
Aariq, line 41: I think this test is a duplicate (of L17-22)  and can be removed

tests/testthat/test-nist.R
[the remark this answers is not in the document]
Aariq: I don't think the user really needs to know the NIST internal IDs.  My opinion is that in this case the CAS should be `NA`.

tests/testthat/test-nist.R
```diff
@@ -129,3 +147,35 @@ test_that("nist_ri works with NAs", {
...
+  expect_true(is.na(test[[2,"RI"]]))
+
+  expect_equal(test[[3,"query"]], "methane")
+  expect_equal(test[[3,"cas"]], "74-82-8")
+  expect_true(is.na(test[[3,"RI"]]))
+
+  expect_equal(test[[4,"query"]], "hexanol")
+  expect_equal(test[[4,"cas"]], "111-27-3")
+  expect_equal(test[[4,"RI"]], 1339)
+})
```
Aariq, line 181: I hate to ask for more tests (because our tests take _so_ long to run already), but it would be nice to see that all the expected columns are there.  So, like, the column names you get from a query with `temp_prog = c("ramp", "isothermal")` should be the column names from a query with `temp_prog = "ramp"` plus the column names from a query with `temp_prog = "isothermal"`.

R/nist.R
```diff
@@ -252,112 +364,75 @@ tidy_ritable <- function(ri_xml) {
...
+    # format name
+    dictionary <- c("alpha" = "\\u03b1",
+                    "beta" = "\\u03b2",
+                    "gamma" = "\\u03b3",
+                    "delta" = "\\U0001d6ff")
+    x <- stringr::str_replace_all(x, dictionary)
+    x
   }
+  sapply(x, format_name)
+}
```
Aariq, line 402: I **love** this, but I think I'd like it to be a little _safer_.  I can't think of a real example, but if there was some compound named something like "deltatalone", this would incorrectly change it to "𝛿talone".  Maybe only do these replacements if they have non-letters next to them?

R/nist.R
```diff
@@ -252,112 +364,75 @@ tidy_ritable <- function(ri_xml) {
...
-  polarity <- match.arg(polarity)
-  temp_prog <- match.arg(temp_prog)
+#' @noRd
+try_url <- function(qurl){
+  try(httr::RETRY("GET",
+                qurl,
+                httr::user_agent(webchem_url()),
+                terminate_on = 404,
+                quiet = TRUE), silent = TRUE)
+}
```
Aariq, line 410: Should this just be in utils.R and get used for _everything_ @stitam?  Maybe as `webchem_try()` to be consistent with other helpers?

R/nist.R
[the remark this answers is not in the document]
Aariq: Definitely don't export this and include `@ noRd`.  Possibly change the name to `format_cas_nist` to avoid confusing future developers?

R/nist.R
```diff
@@ -1,3 +1,128 @@
...
+#'   by searching the NIST WebBook for a corresponding CAS number. If a CAS
+#'   number is found, it will be returned in a \code{tibble} with the
+#'   corresponding information from the NIST retention index database.
+#'
+#' @param query character; the search term
+#' @param from character; type of search term. can be one of \code{"name"},
+#'   \code{"inchi"}, \code{"inchikey"}, or \code{"cas"}. Using an identifier is
+#'   preferred to \code{"name"} since \code{NA} is returned in the event of
+#'   multiple matches to a query. Using an identifier other than a CAS number
+#'   will cause this function to run slower as CAS numbers are used as internal
```
Aariq, line 21: Add something about how "alpha", "beta", etc. will automatically be converted to greek letters with `from = "name"`

R/nist.R
```diff
@@ -1,3 +1,128 @@
...
+    from <- "cas"
+  }
+
+  from <- match.arg(from)
+  type <- match.arg(type, c("kovats", "linear", "alkane", "lee"), several.ok = TRUE)
+  polarity <- match.arg(polarity, c("polar", "non-polar"), several.ok = TRUE)
+  temp_prog <- match.arg(temp_prog, c("isothermal", "ramp", "custom"), several.ok = TRUE)
+
+  if (from == "cas"){
+    query <- format_cas(query)
```
Aariq, line 98: why can't this just be `as.cas(query)`?

R/nist.R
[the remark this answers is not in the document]
Aariq: Also, this appears to do something unexpected with the name of the output. Very likely totally unimportant though.
```
as.cas("78706")
#   78-70-6 
# "78-70-6" 
format_cas("78706")
# 78706.78-70-6 
#     "78-70-6"
```

R/nist.R
```diff
@@ -34,108 +159,145 @@ get_ri_xml <-
...
+        return(dplyr::tibble(query=NA))
+      }
+      if (verbose) message(httr::message_for_status(res))
+      if (res$status_code == 200) {
+        page <- xml2::read_html(res)
+        #Warnings
+        result <- page %>%
+          rvest::html_node("main h1") %>%
+          rvest::html_text()
+        # if cquery not found
```
Aariq, line 183:
```suggestion
        # if query not found
```

R/nist.R
```diff
@@ -34,108 +159,145 @@ get_ri_xml <-
...
+      } else{
+        ri_xml <- lapply(seq_len(nrow(tables)), function(i){
+          ri_xml <- scrape_RI_table(ID, type=tables[i,1], polarity = tables[i,2],
+                          temp_prog = tables[i,3], from=from, verbose=verbose,
+                          cas_found = cas_found)
+          attr(ri_xml, "query") <- query
+          ri_xml
+        })
       }
     }
```
Aariq, line 250: I like this approach!  I didn't get what `tables` was until running it in debug mode, but I see now that you're storing function arguments as rows of a matrix and iterating over rows.  Cool!

R/nist.R
```diff
@@ -170,76 +331,27 @@ tidy_ritable <- function(ri_xml) {
...
+                           "program" = contains("Program"),
+                           "temp" = contains("Temperature (C)"),
+                           "temp_start" = contains("Tstart (C)"),
+                           "temp_end" = contains("Tend (C)"),
+                           "temp_rate" = contains("Heat rate (K/min)"),
+                           "hold_start" = contains("Initial hold (min)"),
+                           "hold_end" = contains("Final hold (min)"),
+                           "reference" = "Reference",
+                           "comment" = "Comment") %>%
+  dplyr::mutate_at(vars(any_of(c("length", "diameter", "thickness", "temp",
```
Aariq, line 352: I _think_ you don't need `vars()` anymore with dplyr 
```suggestion
  dplyr::mutate_at(any_of(c("length", "diameter", "thickness", "temp",
```

Aariq (changes requested):
Got a chance to play around with this PR.  Very nice additions!  I have left some specific comments.

R/nist.R
```diff
@@ -1,3 +1,128 @@
...
+
+  # ri_tables <- purrr::map_dfr(ri_xmls, tidy_ritable, .id = "query") %>%
+  #   dplyr::mutate(query = na_if(query, ".NA"))
+  ri_tables <- dplyr::bind_rows(ri_tables)
+  return(ri_tables)
+}
+
 #' Scrape Retention Indices from NIST
 #'
 #' @param query query of type matching the options in `from`
```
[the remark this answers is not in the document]
ethanbass: Sorry I'm not sure what you're saying should be removed here?

tests/testthat/test-nist.R
ethanbass: good idea. I will add this

tests/testthat/test-nist.R
ethanbass: yes, I noticed this too and almost removed it, but I didn't know if there was a reason. I will remove it now

R/nist.R
ethanbass: the main reason was because I wanted to keep the weird nist_ids in there. I can probably remove the format_cas function if we don't want to do this

R/nist.R
ethanbass: done, though i'm not sure my solution is the best or if there's a better way

tests/testthat/test-nist.R
ethanbass: I added a new test for this: `test_that("nist_ri() works with multiple temperature program arguments"`

R/nist.R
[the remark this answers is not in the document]
Aariq: Oh, nevermind!  I didn't realize this was part of the documentation for the next function in the file.

ethanbass:
Oops, it looks like we inadvertently introduced a couple of typos with the last set of commits. i just pushed a fix

R/nist.R
ethanbass: It looks like the `vars()` actually is still required. I am going to revert this commit

Aariq: `mutate_at()` is superseded, but it's fine to just leave it for now.  The "new" pattern is something like 
```
mutate(across(any_of(c("length", "diameter", "thickness", ...)),  as.numeric))
```

ethanbass:
Ah ok. I’m not really that up on the tidyverse syntax. It seems like it’s
always changing!

Aariq:
No worries.  `mutate_at()` isn't going anywhere anytime soon, so it's fine to leave it.

ethanbass:
Do you guys know why that test is always failing? It seems to work much more reliably on my laptop...

R/nist.R
stitam: I looked at variations across the package: while this structure is quite common, we sometimes use "POST" instead of "GET" and we often have extra arguments, so I don't think we can use this function for everything. We could add an argument to adjust get/post and also add `...` to handle extra arguments, but I'm not sure if this makes it simpler or more complex. I also like the idea, maybe leave it here for now and resolve when we have thought a bit more about this?

Aariq:
CTS is just not very reliable, I think.  I'll re-run the failed tests.

stitam (approved):

PR #384, opened 2022-11-18
fix bug when search matches multiple records

ethanbass:
Sorry, I just found a bug that I introduced with the last pull request. It occurs when a query matches multiple records in the web book. This PR should fix the bug and also adds a test for this scenario.

PR task list:
- [ x] Update NEWS [I don't think it's necessary to update the NEWS for this...?]
- [ x] Add tests (if appropriate)
- [ x] Check package passed

### DIFF
--- a/R/nist.R
+++ b/R/nist.R
@@ -129,8 +129,6 @@ nist_ri <- function(query,
   querynames [is.na(querynames)] <- ".NA"
   names(ri_tables) <- querynames
 
-  # ri_tables <- purrr::map_dfr(ri_xmls, tidy_ritable, .id = "query") %>%
-  #   dplyr::mutate(query = na_if(query, ".NA"))
   ri_tables <- dplyr::bind_rows(ri_tables)
   return(ri_tables)
 }
@@ -146,6 +144,7 @@ nist_ri <- function(query,
 #' @noRd
 #' @import rvest
 #' @import xml2
+#' @importFrom utils URLencode
 #' @return an xml nodeset
 #'
 get_ri_xml <-
@@ -177,7 +176,7 @@ get_ri_xml <-
       ID <- paste0("C", gsub("-", "", query))
     }
     # check for existence of record
-      qurl <- paste0(baseurl, "?", from_str, "=", gsub(" ","+", query), "&Units=SI")
+      qurl <- URLencode(paste0(baseurl, "?", from_str, "=", gsub(" ","+", query), "&Units=SI"))
       webchem_sleep(type = 'scrape')
       if (verbose) webchem_message("query", query, appendLF = FALSE)
       res <- try_url(qurl)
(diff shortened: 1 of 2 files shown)


ethanbass:
I am also proposing to modify the function to replace spaces with `+` in the NIST web book URLs (see e3035167560f494fe5e430c317d0870ca947c144). If you have chemical names with spaces in them (e.g. methyl alcohol), you will currently get an error from `nist_ri`: "Service not available" but these queries work fine from a web browser because the spaces are automatically subbed out with pluses or "%20". (Sorry if this should be a separate pull request).

R/nist.R
```diff
@@ -177,7 +177,7 @@ get_ri_xml <-
       ID <- paste0("C", gsub("-", "", query))
     }
     # check for existence of record
-      qurl <- paste0(baseurl, "?", from_str, "=", query, "&Units=SI")
+      qurl <- paste0(baseurl, "?", from_str, "=", gsub(" ","+", query), "&Units=SI")
```
stitam, line 180: What do you think of `URLencode()` as an alternative?

stitam (changes requested):
Thanks @ethanbass for revisiting this topic! Can you please also add a test where the query contains a space?

R/nist.R
ethanbass: Do you mean this function? https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/URLencode I think that would work too.  So something like:
```
qurl <- URLencode(paste0(baseurl, "?", from_str, "=", query, "&Units=SI"))
```
?

ethanbass: 418bd12edbfcd89c9c5199567ede7d9491df26c6

Aariq: We've used `URLencode` in other functions.  Maybe check and see if it's consistent in webchem and do it whatever way we did it in the past

ethanbass: It seems like it's not that consistent.  Here are some examples I found:

From `opsin_query`:
```
    query_u <- URLencode(query, reserved = TRUE)
```

From `wd_ident`:

```
qurl <- URLencode(qurl)
```

From `get_cid`:'

```
 qurl <- paste0("https://pubchem.ncbi.nlm.nih.gov/rest/pug/",
                         domain, "/",
                         from, "/",
                         "cids/JSON?smiles=",
                         URLencode(as.character(query), reserved = TRUE))
```

From `cir_img`:
```
qurl <- URLencode(paste0(qurl, opts))
```

From `cts_convert`:

```
    query <- URLencode(query, reserved = TRUE)
    from <- URLencode(from, reserved = TRUE)
    to <- URLencode(to, reserved = TRUE)
    baseurl <- "http://cts.fiehnlab.ucdavis.edu/service/convert"
    qurl <- paste0(baseurl, '/', from, '/', to, '/', query)
```
[^ This seems kind of redundant to me. URLencode could be called just once around qurl instead of 3 times].

So ya, it seems like it's not very consistent across the package. I think the way I have it right now seems fine, but happy to go with a different approach too.

Aariq (approved):
Looks good to me!

PR #386, opened 2022-11-29
update to actions/checkout@v3

Aariq:
## Pull Request

I noticed warnings that actions/checkout@v2 is deprecated, this updates to v3

PR task list:
- [ ] Update NEWS
- [ ] Check package passed

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -34,7 +34,7 @@ jobs:
       RSPM: ${{ matrix.config.rspm }}
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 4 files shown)


stitam:
Thanks @Aariq I restarted the tests multiple times and they just wouldn't pass for windows. `rcdk` related errors come up every now and then, do you know why we use `requireNamespace()` in `is.smiles()` and not `@import` or `@importFrom`? I'm wondering if using `@importFrom rcdk parse_smiles` would fix this error?

Aariq:
I think probably `rcdk` is a "Suggests" dependency and not "Imports" because it relies on a Java library, which can be a pain in the ass for install.  There are windows binaries for `rcdk` on CRAN, so it's not clear to me why tests are working on macOS and ubuntu but not windows.  I'll add a `skip_if_not_installed()` to the relevant tests.

stitam (approved):
Seems to work, thanks @Aariq!

Issue #354, opened 2022-03-30
pc_sect() fails when requesting cas for cid 132971


stitam:
I found an example where `pc_sect()` fails unexpectedly.

``` r
webchem::pc_sect(132971, "cas")
#> Error in if (is.character(txt) && length(txt) == 1 && nchar(txt, type = "bytes") < : missing value where TRUE/FALSE needed
```

<sup>Created on 2022-03-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I checked the compound's PubChem page, the CAS registry number is there: https://pubchem.ncbi.nlm.nih.gov/compound/132971

The same query works for other compounds:

``` r
webchem::pc_sect(176, "cas")
#> # A tibble: 21 × 5
#>    CID   Name        Result      SourceName           SourceID  
#>    <chr> <chr>       <chr>       <chr>                <chr>     
#>  1 176   Acetic acid 64-19-7     CAMEO Chemicals      2272      
#>  2 176   Acetic acid 64-19-7     CAMEO Chemicals      9215      
#>  3 176   Acetic acid 64-19-7     CAMEO Chemicals      19328     
#>  4 176   Acetic acid 64-19-7     CAS Common Chemistry 64-19-7   
#>  5 176   Acetic acid 64-19-7     ChemIDplus           0000064197
#>  6 176   Acetic acid 8030-97-5   ChemIDplus           0008030975
#>  7 176   Acetic acid 68475-71-8  ChemIDplus           0068475718
#>  8 176   Acetic acid 99149-56-1  ChemIDplus           0099149561
#>  9 176   Acetic acid 119510-26-8 ChemIDplus           0119510268
#> 10 176   Acetic acid 64-19-7     DrugBank             DB03166   
#> # … with 11 more rows
```

<sup>Created on 2022-03-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Any ideas?

FishParade:
I'm getting the same issue with other CIDs (so far, 8172, 132971). It seems like the error is from the fromJSON function from the jsonlite package:

https://github.com/jeroen/jsonlite/blob/80854359976250f30a86a6992c0d8c3b3d95473d/R/fromJSON.R#L87

As to why only some specific CIDs are causing an error in that line of jsonlite::fromJSON, I'm unsure.

sxthimons:
[...]

PR #392, opened 2023-04-25
Fix pc_sect() failing on certain CIDs

BrianFFish:
Brief description of the PR

Fix for #354. @stitam and @sxthimons confirmed that the cause of pc_sect() sometimes failing with certain CIDs was that certain PubChem records would return NA by httr::content(), and NA as the txt argument in jsonlite::fromJSON() results in an error that halts code execution. Since pc_sect() is a function likely to be run in a loop, it seems better if pc_sect() fails silently on NA returned from httr::content() and continues instead of breaking the loop with an opaque error.

PR task list:
- [ ] Update NEWS
- [ ] Check package passed

### DIFF
--- a/R/pubchem.R
+++ b/R/pubchem.R
@@ -641,6 +641,10 @@ pc_page <- function(id,
     if (verbose) message(httr::message_for_status(res))
     if (res$status_code == 200) {
       cont <- httr::content(res, type = "text", encoding = "UTF-8")
+      # Intercepting any NA cont before it gets to fromJSON.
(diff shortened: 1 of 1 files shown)


stitam:
On second thought, I will fix the tests in a separate PR to keep things more tidy.

Aariq:
Thanks for the fix.  I just wanted to mention that `pc_sect()` is vectorized, so you don't need a for-loop to run it on multiple compounds—just if you want to get data from multiple "sections"

Issue #390, opened 2023-04-24
ChemIDplus has been retired


stitam:
https://www.nlm.nih.gov/pubs/techbull/ja22/ja22_pubchem.html

"On December 12, 2022, [PubChem](https://pubchem.ncbi.nlm.nih.gov/) will serve as NLM's single source for chemical information. NLM is retiring ChemIDplus and the Drug Information Portal, two chemical property information sites, to better focus our development efforts on a single, integrated source of chemical information. All of the data found in ChemIDplus and the Drug Information Portal is currently available and will continue to be available in [PubChem](https://pubchem.ncbi.nlm.nih.gov/)."

In webhem, `ping_service("ci")` still returns `TRUE` but the `ci_query()` now returns a list of `NA`-s regardless of the input.

Since the webservice is no longer working, `ci_query()` should be removed from the package, along with tests, references, etc.

Aariq:
If possible, this should deprecated, not removed.  Ideally it would continue to work (by passing args to the appropriate pc_* function) and print a warning with information about how to update code.  Then in a later version, it would print an error with information about how to update code.

We should probably eventually develop some kind of deprecation policies since this kind of thing will continue to happen as APIs evolve.

stitam:
Thanks @Aariq, valid point about deprecating first instead of "defunct"-ing in one step. 

I understand that normally we would remove functions over time, but in this case it is not webchem that changed but the webservice, so this might be a different scenario? I haven't looked into your suggestion in detail but probably some of the list elements could be reproduced through `pc_sect()` calls and some downstream processing.

On the other hand, I do not have a recorded response to test any new implementation with. Also  I'm a little worried about the amount of effort it would require from us to make it work again (another difference from normal function lifecycle), not sure whether this should be our problem?

In general, if APIs stop working then in most cases we won't be able to fix it, so maybe our role is just to make sure webchem is not polluted with functions that no longer work? This one is different because it no longer works, but luckily we already implemented PubChem so we might be able to rewrite the backend to emulate the user experience. Just not sure if we should put effort into something that's already outdated.

Aariq:
We could just defunct it and print an error that blames the API provider and a suggestion for an alternative.  I'd be ok with that.  Whatever we decide, it might be good to take some notes about it in CONTRIBUTING.md

stitam:
Oh, that's exactly what I had in mind, defunct with an informative message! Great, I'll open the PR in a bit.

PR #393, opened 2023-04-25
Defunct ci_query() and purge ChemIDplus from package

stitam:
This PR is related to issue #390.

The changes turn `ci_query()` defunct with an informative message and remove all traces of ChemIDplus from the package.

Tests won't pass because other tests are still failing, but to keep things tidy, I'd rather fix those in a separate PR.

PR task list:
- [x] Update NEWS
- [ ] Check package passed


Aariq (approved):

Aariq:
I know I approved this and it's already merged, but just remembered you should bump the version.  I think this would be a minor version since it's more than just a bug fix?  So should be on 1.3.0 now I think?

stitam:
Sure, let's wrap up and submit a minor release!

PR #398, opened 2023-05-08
Reformat cas

ethanbass:
- CAS numbers are automatically reformatted using `as.cas` in bcpc, cts, etox, flavornet, pan, pubchem, srs functions.
- Adds `verbose` argument to `as.cas`.
- Changes the behavior of `as.cas` so it returns a named character vector whose names are the character strings from the original query.
- Alters test for `as.cas` (since it now retains the names of the original query).
- This pull request also changes the way names are assigned to `lapply` output because I ran into a problem with the way the output was being formatted for "non-cas" cas numbers like "balloon".
- Updated NEWS

PR task list:
- [ x] Update NEWS
- [ x] Add tests (if appropriate) [I didn't add any tests here, but let me know if you want me to add additional tests (e.g. that CAS numbers without dashes are being correctly reformatted).
- [x ] Update documentation with `devtools::document()`
- [ x] Check package passed

### DIFF
--- a/R/bcpc.R
+++ b/R/bcpc.R
@@ -47,8 +47,11 @@ bcpc_query <- function(query, from = c("name", "cas"),
     message('To search by compound name use "name" instead of "commonname"')
     from <- "name"
   }
+
   from <- match.arg(from)
   bcpc_idx <- build_bcpc_idx(verbose, ...)
+
+  names(query) <- query
 
   foo <- function(query, from, verbose) {
     if (is.na(query)) {
@@ -64,6 +67,7 @@ bcpc_query <- function(query, from = c("name", "cas"),
     }
 
     if (from == "cas") {
+      query <- as.cas(query, verbose = verbose)
       names <- bcpc_idx$names[bcpc_idx$source == "rn"]
       # select only first link
       links <- bcpc_idx$links[bcpc_idx$source == "rn"]
@@ -164,7 +168,6 @@ bcpc_query <- function(query, from = c("name", "cas"),
     }
   }
   out <- lapply(query, function(x) foo(x, from = from, verbose = verbose))
-  names(out) <- query
   class(out) <- c("bcpc_query", "list")
   return(out)
 }
--- a/R/bcpc.R
+++ b/R/bcpc.R
@@ -54,6 +54,14 @@ bcpc_query <- function(query, from = c("name", "cas"),
   names(query) <- query
 
   foo <- function(query, from, verbose) {
+    if (from == "cas") {
+      query <- as.cas(query, verbose = verbose)
+      names <- bcpc_idx$names[bcpc_idx$source == "rn"]
+      # select only first link
+      links <- bcpc_idx$links[bcpc_idx$source == "rn"]
+      linknames <- bcpc_idx$linknames[bcpc_idx$source == "rn"]
+      cname <-  linknames[tolower(names) == tolower(query)]
+    }
     if (is.na(query)) {
       if (verbose) webchem_message("na")
       return(NA)
@@ -64,15 +72,6 @@ bcpc_query <- function(query, from = c("name", "cas"),
       links <- bcpc_idx$links[bcpc_idx$source == "cn"]
       names <- bcpc_idx$linknames[bcpc_idx$source == "cn"]
       cname <-  query
-    }
-
-    if (from == "cas") {
-      query <- as.cas(query, verbose = verbose)
-      names <- bcpc_idx$names[bcpc_idx$source == "rn"]
-      # select only first link
-      links <- bcpc_idx$links[bcpc_idx$source == "rn"]
-      linknames <- bcpc_idx$linknames[bcpc_idx$source == "rn"]
-      cname <-  linknames[tolower(names) == tolower(query)]
     }
 
     takelink <- links[tolower(names) == tolower(query)]
--- a/R/cts.R
+++ b/R/cts.R
@@ -45,9 +45,6 @@ cts_compinfo <- function(query, from = "inchikey",
       if (verbose) webchem_message("na")
       return(NA)
     }
-    if (from == "cas"){
-      query <- as.cas(query, verbose = verbose)
-    }
     if (verbose) webchem_message("query", query, appendLF = FALSE)
     if (!is.inchikey(query, verbose = FALSE)) {
       if (verbose) message("Input is not a valid inchikey.")
@@ -162,12 +159,12 @@ cts_convert <- function(query,
   }
 
   foo <- function(query, from, to, first, verbose){
+    if (from == "cas"){
+      query <- as.cas(query, verbose = verbose)
+    }
     if (is.na(query)) {
       if (verbose) webchem_message("na")
       return(NA)
-    }
-    if (from == "CAS"){
-      query <- as.cas(query, verbose = verbose)
     }
     if (verbose) webchem_message("query", query, appendLF = FALSE)
     query <- URLencode(query, reserved = TRUE)
--- a/R/cts.R
+++ b/R/cts.R
@@ -167,7 +167,7 @@ cts_convert <- function(query,
       return(NA)
     }
     if (from == "CAS"){
-      query <- sapply(query, as.cas, verbose = verbose)
+      query <- as.cas(query, verbose = verbose)
     }
     if (verbose) webchem_message("query", query, appendLF = FALSE)
     query <- URLencode(query, reserved = TRUE)
--- a/R/cts.R
+++ b/R/cts.R
@@ -34,7 +34,7 @@ cts_compinfo <- function(query, from = "inchikey",
                          verbose = getOption("verbose"), inchikey){
 
   if (!ping_service("cts")) stop(webchem_message("service_down"))
-
+  names(query) <- query
   if (!missing(inchikey)) {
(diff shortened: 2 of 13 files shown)


R/flavornet.R
```diff
@@ -34,13 +34,17 @@ fn_percept <- function(query, from = "cas", verbose = getOption("verbose"),
...
   match.arg(from)
+  names(query) <- query
   foo <- function (query, verbose){
     if (is.na(query)) {
       if (verbose) webchem_message("na")
       return(NA)
     }
+    if (from == "cas"){
+      query <- as.cas(query, verbose = verbose)
+    }
```
Aariq, line 45: I think this should be moved up above line 39 so in the case that the CAS number isn't valid, and `as.cas()` returns `NA`, the website isn't queried

R/bcpc.R
```diff
@@ -64,6 +67,7 @@ bcpc_query <- function(query, from = c("name", "cas"),
     }
 
     if (from == "cas") {
+      query <- as.cas(query, verbose = verbose)
```
Aariq, line 70: It might be good to move this (or the whole `if (from == "cas")`) above line 57 so when the CAS number isn't valid and returns NA, the API isn't queried

R/cts.R
```diff
@@ -45,7 +45,10 @@ cts_compinfo <- function(query, from = "inchikey",
       if (verbose) webchem_message("na")
       return(NA)
     }
-    if(verbose) webchem_message("query", query, appendLF = FALSE)
+    if (from == "cas"){
```
Aariq, line 48: Unless I'm mistaken, `cts_compinfo()` doesn't have an option for CAS numbers.  Anything other than `from = "inchikey"` should error.

Aariq, line 48: Unless I'm mistaken, `cts_compinfo()` doesn't have an option for CAS numbers.  Anything other than `from = "inchikey"` should error.

R/cts.R
```diff
@@ -163,7 +166,10 @@ cts_convert <- function(query,
       if (verbose) webchem_message("na")
       return(NA)
     }
-    if(verbose) webchem_message("query", query, appendLF = FALSE)
+    if (from == "CAS"){
```
Aariq, line 169: Should this be lowercase?
```suggestion
    if (from == "cas"){
```

R/etox.R
```diff
@@ -46,6 +46,8 @@ get_etoxid <- function(query,
                        verbose = getOption("verbose")) {
 
   if (!ping_service("etox")) stop(webchem_message("service_down"))
+  names(query) <- query
```
Aariq, line 49: The output of `get_etoxid` contains the formatted CAS number, not the original query, in the case of a CAS number without hyphens.  Should it be the original query for consistency with other functions?

R/pubchem.R
```diff
@@ -133,6 +133,9 @@ get_cid <-
   }
     #input validation
     from <- tolower(from)
+    if (from == "cas"){
```
Aariq, line 136: "cas" isn't one of the options for `get_cid`, but maybe it should be.  Currently, you supply a CAS number with `from = "xref/rn"`.  You could make this work with the function as-is or add `"cas"` as an alias.

Aariq (changes requested):
There are a few problems that need fixing, but the concept is great.

R/cts.R
ethanbass: yes, it should. I got confused because I saw the upper case "CAS" in the documentation, but I see now that these arguments are converted to lower case inside the `match.arg`.

R/flavornet.R
ethanbass: good point.

R/cts.R
ethanbass: oops. you're right! i'll remove this

R/bcpc.R
ethanbass: good point. I moved the whole thing up. Does that seem OK to you?

R/etox.R
ethanbass: I think this is generally true with the way i've implemented this, since i am overwriting the original query with the reformatted CAS numbers. (This is also the way that `nist_ri` is working in the current version). I'm not sure what would be a better alternative here?

R/pubchem.R
ethanbass: good point. I fixed this and added an "alias". I'm not sure if this needs to be documented? I'm finding the documentation for get_cid kind of confusing TBH, so I'm not really sure where this would go.

ethanbass:
> Thanks @ethanbass! Wondering if you folks want this to be part of the next release? I need to resubmit v1.3.0 to CRAN with some comments so I can wait until we merge this PR.

Whatever you all want to do in this regard is fine with me. I'm not really familiar with your release schedule

R/cts.R
```diff
@@ -159,11 +159,14 @@ cts_convert <- function(query,
   }
 
   foo <- function(query, from, to, first, verbose){
+    if (from == "cas"){
+      query <- as.cas(query, verbose = verbose)
+    }
     if (is.na(query)) {
       if (verbose) webchem_message("na")
       return(NA)
     }
```
Aariq, line 168: Invalid cas numbers still aren't making it here for some reason,  e.g.

``` r
cts_convert('111111', "cas", "chemical name", verbose = TRUE)
# 111-11-1: Checksum is not correct! 5 vs. 1
# Querying NA. Error in strsplit(URL, "") : non-character argument

cts_convert(as.cas('111111'), "cas", "chemical name", verbose = TRUE)
# Query is NA. Returning NA.
# $<NA>
# [1] NA
```

The output of these two should be identical.  Might be good to write a test for this.

R/pubchem.R
```diff
@@ -133,6 +133,10 @@ get_cid <-
   }
     #input validation
     from <- tolower(from)
+    from <- ifelse(from == "cas", "xref/rn", from)
+    if (from == "xref/rn"){
+       query <- as.cas(query, verbose = verbose)
+    }
```
Aariq, line 139: The other option, which I'm not sure is better, is to start a separate section with a `case_when()` or `switch()` or something to handle this, making it extendable to other "aliases".  I think I like the simpler solution for now.

Aariq (changes requested):

R/etox.R
Aariq: `get_etoxid()` returns the formatted query
``` r
get_etoxid("34123596", from = "cas") # Isoproturon
## A tibble: 1 × 3
#  query      match       etoxid
#  <chr>      <chr>       <chr> 
# 1 34123-59-6 Isoproturon 7442  
```
But the way you've implemented this in the other functions, the original query is part of the returned object.  E.g. 

``` r
fn_percept("34123596")
# 34123596 
#      NA 
```
Should the output tibble have `"34123596"` or `"34123-59-6"` in the `query` column?

R/pubchem.R
Aariq: I would maybe edit the `@details` section (starting line 18) so that the rendered help file looks something like:

>Details
>Valid values for the from argument depend on the domain:
>- compound: "name", "smiles", "inchi", "inchikey", "formula", "sdf", "cas" (an alias for \"xref/RN"), \<xref\>, \<structure search\>, \<fast search\>.

R/pubchem.R
Aariq: Yeah, I think its good as is.  The more I think about it, it seems unlikely there will be a need for many other "aliases"

R/etox.R
ethanbass: we can overwrite the query using the `.id` argument in `bind_rows` (or add an additional column with the original query?): e.g. `out <- dplyr::bind_rows(out, .id = "query")`. I guess the question is what would be most desirable here?

R/pubchem.R
ethanbass: done

R/cts.R
ethanbass: The only difference here for me is the name is NA in the second example, but I think this is the expected behavior?

ethanbass: Or do we want to somehow preserve the names here (if queries are given as a named character vector?) I guess there should be a way to do this, but it would complicate things.

Aariq: Ah, I must have been using an old commit or something.  Works now!

R/etox.R
Aariq: I would suggest just having the one `query` column and it should be whatever the user supplied (before `as.cas()`)

stitam: I agree with keeping the original query: some users may want to use this variable for merging data; if we convert the queries automatically we may add unnecessary frustration. Whoever needs to convert the query column afterwards can always use `as.cas()`.

ethanbass: OK sounds good. this is done in 56e357f

ethanbass:
Do we want to add additional tests associated with this PR?

Aariq (approved):

stitam (approved):

stitam:
Not sure why checks on ubuntu-20.04 (devel) passed before and do not pass now, maybe we should update GitHub actions? A possible lead here: https://community.rstudio.com/t/github-action-failure-with-rcmd-check-on-ubuntu-devel/129727. I'm going to merge this PR and fix the check afterwards.

Issue #355, opened 2022-04-01
Update use cases


Aariq:
Might be good to add some more recent use cases to the README.  E.g.:

https://doi.org/10.1080/08958378.2022.2051646

Aariq:
Could be done dynamically/automatically as well with a GH action that uses `rcrossref`?

PR #402, opened 2023-06-02
add DOI badge (#347) and additional use cases (#355) to README

### LINKED ISSUE #347: Add DOI badge to README
We should add a DOI badge to the README that links to the webchem paper

Aariq:
Just adds DOI badge and additional use cases to README.  I found the additional use cases through researchgate.net and they are by no means comprehensive.  I just thought it would be nice to have some recent papers that used webchem.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -77,16 +77,7 @@ install_github("ropensci/webchem")
 
 ### Use Cases
 
-Have you used `webchem` in your work?  Please let us know by opening an issue or making a pull request to edit this section!
-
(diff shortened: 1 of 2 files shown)


stitam (commented):
Thanks @Aariq! I am wondering if instead of adding use cases manually we should just add e.g. a google scholar link? https://scholar.google.com/scholar?oi=bibs&hl=en&cites=14244442030948237605&as_sdt=5. What do you think?

Aariq:
That sounds fine to me, although not everyone that cites webchem is _using_ webchem (includes reviews and such as well).  I'll update the PR accordingly.

stitam (approved):

Issue #399, opened 2023-05-09
`pan_query()` still broken.  Should it be made defunct for v 1.3.0?


Aariq:
`pan_query()` returns a message that it is broken when used.  I couldn't find an open issue or PR fixing it. Should we just mark it as defunct in the next version (#395)?

stitam:
This is different from ChemIDPlus because the website is still there we just cannot access it. I asked the organisation for clarification. If there is an API we can rewrite the function if there isn't and they prohibit programmatic access then we need to defunct it. Thanks @Aariq, I agree, let's sort this out before v1.3.0.

Aariq:
Ah I remember now.  I couldn't find the issue where we discussed this previously.

Aariq:
Sad to see it go, but I agree

stitam:
Forgot to remove PAN from README.

PR #404, opened 2023-06-08
PAN

stitam:
Related to Issue #399.

This PR removes PAN from the DESCRIPTION file and the vignette. In the vignette, `pan_query()` was used to retrieve WHO acute toxicity data. This is now replaced with the number of aromatic rings, using `chembl_query()`.

I still see PAN:

* in `codemeta.json` but that will be updated during prep for new release. 
* in NAMESPACE but that's because the defunct function still has to be exported.
* in webcehm-defunct.R, obviously.

Other than these, I think we're done. @Aariq do you see PAN anywhere else? Thanks :)

PR task list:
- [x] Check package passed


Aariq (approved):
Looks good to me!  I didn't find any other references to PAN and the changes to the vignette look good.

Issue #407, opened 2023-07-05
Character CID that can't be coerced to numeric in `pc_prop()`  results in `NA` output for all CIDs


schymane:
We have been using the PubChem download files (from various sources within PubChem) to create CID lists, which we then feed into webchem to obtain other properties of interest. Today one CID in a big list was NULL, resulting in a very surprising (and small) "NA" output without any indication of the reason for failure. It took me a while to find the cause!
I have now updated my scripts to remove the NULL in advance (I already removed NAs), but perhaps it's worth adding a check within the function to screen the CID list in advance and print a warning? I just grepped for "NULL". 

Here's a much smaller example to demonstrate (I had a list of 6K CIDs originally)

```
library(webchem)
library(data.table)
selected_properties <- c("MolecularFormula","ExactMass","XlogP",
                         "CanonicalSMILES","IsomericSMILES",
                         "InChI","InChIKey","Title","IUPACName")
test_cids <- c("1234","NULL")
as.data.table(webchem::pc_prop(test_cids, selected_properties))
# returns NA (see screenshot)
as.data.table(webchem::pc_prop(c("1234"), selected_properties))
# returns the expected result (see screenshot)
```

![image](https://github.com/ropensci/webchem/assets/4070141/5f2e0f34-5b57-4b49-89af-9027225091cc)

Thanks,
Emma

Aariq:
I'm not able to reproduce this with the dev version of webchem with `NULL` or `NA`, but I do get it with `Inf` or `"baloon"` (E.g.)

``` r
library(webchem)
selected_properties <- c("MolecularFormula","ExactMass","XlogP",
                         "CanonicalSMILES","IsomericSMILES",
                         "InChI","InChIKey","Title","IUPACName")
pc_prop(c("1234", NA), selected_properties)
#>    CID MolecularFormula
#> 1 1234       C28H40N2O5
#> 2   NA             <NA>
#>                                                    CanonicalSMILES
#> 1 CC(C)C(CCCN(C)CCC1=CC(=C(C=C1)OC)OC)(C#N)C2=CC(=C(C(=C2)OC)OC)OC
#> 2                                                             <NA>
#>                                                     IsomericSMILES
#> 1 CC(C)C(CCCN(C)CCC1=CC(=C(C=C1)OC)OC)(C#N)C2=CC(=C(C(=C2)OC)OC)OC
#> 2                                                             <NA>
#>                                                                                                                                                         InChI
[...]

PR #408, opened 2023-07-07
Ensure all CIDs are positive integers when using pc_prop()

stitam:
Related to Issue #407.

This PR updates `pc_prop()` such that:

* All queries are converted to positive integers
* If this produces `NA` values, the function prints informative messages, if verbose messages are enabled
* Column `CID` of the final output table is the original vector of queries.

The last point  is to ensure downstream processing e.g. merging tables is not impacted.

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/R/pubchem.R
+++ b/R/pubchem.R
@@ -294,7 +294,8 @@ get_cid <-
 #' @param verbose logical; should a verbose output be printed to the console?
 #' @param ... currently not used.
 #'
-#' @return a data.frame
+#' @return a tibble; each row is a queried CID, each column is a requested
+#' property.
 #' @seealso \code{\link{get_cid}}, \code{\link{pc_sect}}
 #' @references Wang, Y., J. Xiao, T. O. Suzek, et al. 2009 PubChem: A Public
 #' Information System for
@@ -432,8 +433,9 @@ pc_prop <- function(cid, properties = NULL, verbose = getOption("verbose"), ...)
         }
       }}
     rownames(out) <- NULL
-    class(out) <- c("pc_prop", "data.frame")
     out$CID <- cid_o
+    out <- tibble::as_tibble(out)
+    class(out) <- c("pc_prop", class(out))
     return(out)
   }
   else {
--- a/R/pubchem.R
+++ b/R/pubchem.R
@@ -334,12 +334,39 @@ pc_prop <- function(cid, properties = NULL, verbose = getOption("verbose"), ...)
 
   if (!ping_service("pc")) stop(webchem_message("service_down"))
 
+  cid_o <- cid
+
+  if (verbose) message("Coercing queries to positive integers. ", appendLF = FALSE)
+
+  cid <- suppressWarnings(as.integer(cid))
+
+  if (verbose) {
+    index <- which(is.na(cid) & !is.na(cid_o))
+    if (length(index) > 0) {
+      for (i in index) {
+        message(paste0(cid_o[index], " coerced to NA. "), appendLF = FALSE)
+      }
+    }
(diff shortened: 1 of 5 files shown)


R/pubchem.R
```diff
@@ -334,12 +334,39 @@ pc_prop <- function(cid, properties = NULL, verbose = getOption("verbose"), ...)
...
+
+  if (verbose) message("Coercing queries to positive integers. ", appendLF = FALSE)
+
+  cid <- suppressWarnings(as.integer(cid))
+
+  if (verbose) {
+    index <- which(is.na(cid) & !is.na(cid_o))
+    if (length(index) > 0) {
+      for (i in index) {
+        message(paste0(cid_o[index], " coerced to NA. "), appendLF = FALSE)
```
Aariq, line 347: I like this approach!

R/pubchem.R
```diff
@@ -334,12 +334,39 @@ pc_prop <- function(cid, properties = NULL, verbose = getOption("verbose"), ...)
 
   if (!ping_service("pc")) stop(webchem_message("service_down"))
 
+  cid_o <- cid
```
Aariq, line 337: Can you edit the documentation to say something about what is considered a valid `cid`?  Right now it says it has to be character, which isn't true.

R/pubchem.R
```diff
@@ -334,12 +334,39 @@ pc_prop <- function(cid, properties = NULL, verbose = getOption("verbose"), ...)
...
+
+  if (verbose) message("Coercing queries to positive integers. ", appendLF = FALSE)
```
Aariq, line 339: You could add a simple test using `verbose = TRUE` to keep test coverage up

tests/testthat/test-chembl.R
```diff
@@ -14,7 +14,7 @@ test_that("chembl_query()", {
   o4 <- chembl_query("CHEMBL771355", resource = "assay")
 
   expect_type(o1, "list")
-  expect_equal(length(o1[[1]]), 34)
+  expect_equal(length(o1[[1]]), 35)
```
Aariq, line 17: What happened here?  API change?

tests/testthat/test-pubchem.R
```diff
@@ -105,6 +105,12 @@ test_that("pc_prop", {
   c <- pc_prop("5564", properties = c("CanonicalSmiles", "InChiKey"))
   expect_true(is.na(b))
```
Aariq, line 106: Should this return a tibble for type consistency?

tests/testthat/test-pubchem.R
```diff
@@ -105,6 +105,12 @@ test_that("pc_prop", {
...
   expect_equal(ncol(c), 3)
+
+  cids <- c(5564, NA, -1, "balloon", "NULL", "2244")
```
Aariq, line 109: I would actually leave `"NULL"` out of the test because it is confusing and the same as `"balloon"`

Aariq (approved):
Suggested some minor changes but went ahead and approved so as to not hold you up

tests/testthat/test-chembl.R
stitam: Not sure I'm afraid. Response looks similar, but there is an extra element in the returned list. I don't know which element is the extra; maybe we should look these up manually so next time we know what the change was?

tests/testthat/test-pubchem.R
stitam: Sure, as we're already fiddling with the function it's probably a good idea to upgrade class as well.

PR #414, opened 2024-02-16
update actions/checkout

Aariq:
GitHub has started complaining about actions that use old versions of node, including actions/checkout v3, so this updates to v4


stitam (approved):
Thanks @Aariq!

Issue #295, opened 2020-09-29
improve aw_query()


andschar:
`aw_query()` currently returns a list of lists of vectors, when it finds a match and a vector `NA` of length 1, if it doesn't. In my opinion this is quite cumbersome to wrangle with in R. Example:

```r
todo = c("630-20-6", "1070-78-6", "1071-83-6", "79-34-5")
aw = webchem::aw_query(todo, from = 'cas')
```
The first two entries are `NA` vectors with length 1, the second two entries are lists of vectors. In my opinion it would be better to return a tibble if something is found, because one can use `data.table::rbindlist()` or `dplyr::bind_rows()` then. E.g.: `data.table::rbindlist(aw[ !is.na(aw) ])`. Though, we could also think about returning `tibble::tibble(query = '630-20-6')` in the case of no match instead of `NA`, meaning one wouldn't have to exclude `NA` explicitly when binding. AW returns tabular-freinldy data, so I think it's quite easy to implement the tibbles.

This also belongs to #218.

Also, `aw_query()` results have the class attribute `aw_query` assigned. Designed by @EDiLD to work with the `S3`-OO system. Should we continue using this? Lately, we haven't made use of any OO-functions. I think we haven't really agreed on that up to now. Also to discuss further in #228.

Aariq:
Just wanted to note that this still applies to `bcpc_query()`

PR #420, opened 2024-11-07
Feature bcpc extend features

AlexisBrgn:
## Pull Request

Brief description of the PR

Extended features from bcpc.R : 

- Added locales to search from : Chinese, French, Russian
- Added identifier to search from : inchikey
- Refactored to easily add locales and identifier to search from.
- `bcpc_query` now returns locales

This PR is unsolicited but it was a shame the exposed functions didn't allow to search in other languages when the info was right there. I don't like refactoring other people's code but the existing codebase would've made extending features a bit messy.
Maybe this is the occasion to address #295? I'm not sure it understand it fully to fix.

PR task list:
- [x] Update NEWS
- [ ] Unit test extracted functions from the refactoring?
- [ ] Check package passed
- [x] test-bcpc.R passed

### DIFF
--- a/R/bcpc.R
+++ b/R/bcpc.R
@@ -32,7 +32,7 @@
 #' # use CAS-numbers
 #' bcpc_query("79622-59-6", from = 'cas')
 #' }
-bcpc_query <- function(query, from = c("name", "cas"),
+bcpc_query <- function(query, from = c("name", "cas", "inchikey", "name_fr"),
                        verbose = getOption("verbose"),
                        type, ...) {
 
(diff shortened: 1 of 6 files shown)


stitam:
Thanks @AlexisBrgn for opening this PR. Some of the webservices changed which broke a few tests we already had. I fixed these, can you please rebase on the current master? After rebase, please also rerun devtools::document(), this will remove man/webchem.Rd.

Is this PR still work in progress?

AlexisBrgn:
Thanks for the heads up I'll dig into it as soon as I manage to free up some time to contribute.

 > Is this PR still work in progress?

Yes it is still very rough, the basic features work!

Aariq:
#295 is basically saying we should return tibbles whenever possible rather than nested lists, if I remember correctly.

Aariq:
I know this is still a WIP, but could you update the documentation to include info on the new `from` options when you get a chance?  That would make it easier to give feedback on this PR.

R/bcpc.R
```diff
@@ -183,71 +69,280 @@ bcpc_query <- function(query, from = c("name", "cas"),
...
+    idx_sources <- data.frame(
+      name = c("rn", "inchikey", "cn", "fr", "ru", "zh"),
+      code = c(TRUE, TRUE, FALSE, FALSE, FALSE, FALSE),
+      url = c(
+        "https://pesticidecompendium.bcpc.org/index_rn.html",
+        "www.bcpcpesticidecompendium.org/index-inchikey.html",
+        "https://pesticidecompendium.bcpc.org/index_cn.html",
+        "https://pesticidecompendium.bcpc.org/index-fr.html",
+        "https://pesticidecompendium.bcpc.org/index-ru.html",
+        "https://pesticidecompendium.bcpc.org/index-zh.html"
```
Aariq, line 93: These URLs all appear to re-direct. E.g. http://www.bcpcpesticidecompendium.org/index-inchikey-frame.html, http://www.bcpcpesticidecompendium.org/index-zh-frame.html

stitam:
> #295 is basically saying we should return tibbles whenever possible rather than nested lists, if I remember correctly.

Yes, I also think the issue was about returning tibbles. Mote than 4 years later I still think it's a good idea :) Without looking at the code in the PR I would think tibbles may be out of scope for this PR but if it's already done, great :)

R/bcpc.R
AlexisBrgn: I admit I blindly pasted the format from the previous code.
After a bit of digging I found these url actually redirect to [http://www.bcpcpesticidecompendium.org](http://www.bcpcpesticidecompendium.org).
For instance https://pesticidecompendium.bcpc.org/index_rn.html -[302]-> http://www.bcpcpesticidecompendium.org/index_rn.html
However the "redirection" to a *-frame.html is initiated by `http://www.bcpcpesticidecompendium.org/index_{from}.html` in js `if (top==self) self.location.href="index_rn_frame.html";` and should not happen when the code executes.
I will therefore replace all the URL to the format `http://www.bcpcpesticidecompendium.org/index_{from}.html`.

AlexisBrgn:
> > #295 is basically saying we should return tibbles whenever possible rather than nested lists, if I remember correctly.
> 
> Yes, I also think the issue was about returning tibbles. Mote than 4 years later I still think it's a good idea :) Without looking at the code in the PR I would think tibbles may be out of scope for this PR but if it's already done, great :)

Looking at the code I think it is out of scope too, but I'm willing to tackle the issue in another PR.

PR #440, opened 2025-10-29
Let bcpc_query also find derivatives

### LINKED ISSUE #439: Query for derivatives in `bcpc_query`
Some active substances like 2,4-D [1] that have an ISO common name list a number of "derivatives", like 2,4-D-meptyl [2] (salts or esters of acids or alcohols). Under "Approval" the pages for the derivatives say "Parent-ISO".

As these substances are actual ingredients of plant protection products, I would like to use the unique names listed by the BCPC for the derivatives and make them searchable using `bcpc_query`. These names are based on certain recommendations listed at the site [3]. Unfortunately it seems they can be ambiguous, but I assume that the nomenclature used for such derivatives at the BCPC site is consistent, which would make it suitable for canonisation of such names. There is an index of the derivatives that one can study [4].

At first glance, it seems the HTML is pretty much the same as for the active substances so it seems it could be possible to return an equivalent set of return values if the derivatives would be included in the search.

[1] http://www.bcpcpesticidecompendium.org/2,4-d.html
[2] http://www.bcpcpesticidecompendium.org/derivatives/2,4-d-meptyl.html
[3] http://www.bcpcpesticidecompendium.org/esters.html
[4] http://www.bcpcpesticidecompendium.org/derivatives/index-derivatives-frame.html

jranke:
Closes #439

Brief description of the PR

PR task list:
- [x] Update NEWS
- [x] Check package passed

Regarding the last item, I did check if the package passed, but I got the same four test failures for `test_chembl` that occur in CI on the master branch.


stitam:
Thanks @jranke for opening this PR, I'll rebase and then we'll see if the checks pass.

stitam (approved):
Thanks @jranke the PR looks good. I'll merge in a bit.

Issue #449, opened 2025-12-01
Get ChEMBL webservice schema


stitam:
When I make a webservice query to ChEMBL some fields in the response are often NULL. Because the offline version of ChEMBL contains NA in these fields, and it is generally easier to work with NA-s in R, it makes sense to convert NULL-s to NA-s. But do we convert to NA_character_ or e.g. NA_integer_? It depends on the field, but the webservice query will not have this information. The solution for this is to somehow know which field has what class and assign the appropriate NA type.

stitam:
This issue is fixed. When we use `chembl_query()` with `mode = "ws"` (webservice), the function pulls `schema.json` for the resource, caches it for the session, and applies the rules. Overall, we retrieve schema files once per session per resource, which I think is OK. I am  closing this issue.

PR #455, opened 2026-04-27
Expand ChEMBL resources and standardize webservice output

stitam:
Related to issue #449.

* Implemented several new ChEMBL resources: `assay`, `drug_indication`, `drug_warning`, `go_slim`.
* Streamlined internal handling of resource schemas and their enforcement.
* When using `chembl_query()` with `mode = "ws"` (web service), webchem retrieves and caches the schema for the requested ChEMBL resource during the session. The schema is then used to standardize the output. For example, if a field is defined as a "string", the returned value is converted to a character vector.
* Schema enforcement also improves handling of missing values. The web service may return `NULL` or empty lists for missing fields, which complicates downstream processing. Using the schema, these are converted to schema-appropriate `NA` values. This behavior is now hard-coded and cannot be disabled.
* Another benefit is improved comparison between online and offline results. We now use  `all.equal()` to check whether web service and offline responses match.
* In some cases, the web service response is slightly inconsistent with the schema (e.g., fields missing from the schema). When this occurs, `chembl_query()` issues warnings and leaves those fields unchanged.

Note, I disabled some CTS tests. It seems to be a webservice issue. I contacted the maintainers.

PR task list:
- [x] Update NEWS
- [x] Check package passed

### DIFF
--- a/.github/workflows/check-pkgdown.yml
+++ b/.github/workflows/check-pkgdown.yml
@@ -1,3 +1,5 @@
+# based on https://docs.ropensci.org/rotemplate/
+
 on:
     push:
       branches:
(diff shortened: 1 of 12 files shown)


stitam:
Some checks for pubchem (not chembl) kept failing on mac. It seemed this was related to timeout on queries, so I did some refactoring to make the requests more robust. These did not fix the issue at the time. However, the tests pass now, so I'll just merge. We may have to record more tests later.

PR #466, opened 2026-07-28
add action with `RUN_ONLINE_TESTS: true`

Aariq:
This adds a GitHub action that runs on the 1st of the month at 3:00am or when you click a button (see screenshot below) that has `RUN_ONLINE_TESTS` set to `"true"` so as to not skip tests that hit APIs. The idea is that once a month (or whenever you want), you get a notification if any tests fail, then you can investigate manually.  You can also run this manually on any PR branch.  It still runs on all platforms in case there are platform-specific problems with the R code of these functions.

Screenshot 2026-07-28 at 10 36 05 AM

### DIFF
--- a/.github/workflows/R-CMD-check-online-tests.yaml
+++ b/.github/workflows/R-CMD-check-online-tests.yaml
@@ -6,7 +6,7 @@ on:
     - cron: '0 3 1 * *'
   workflow_dispatch:
 
-name: R-CMD-check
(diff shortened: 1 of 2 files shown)


stitam:
Thanks @Aariq for adding this action. I haven't used schedules before, cool.

Aariq:
FYI I triggered it manually just now, just to test that it works

